PR #16, opened 2021-10-10
Don't use internals that are not accessible in J9+

Geolykt:


lokka30:
Thank you very much - it seemed I used the wrong import. :)

Issue #17, opened 2021-10-10
Improve .gitignore


MrNemo64:
The .gitifnore file is an important part of a git repository, how ever the .gitignore of this repository doesn't cover all files. Since it only contemplates Intell Idea, contributors using other ides will have trouble. I propouse using a more extense file, there are good generators online that can do the work, for example, [this ](https://www.toptal.com/developers/gitignore/api/eclipse,intellij,maven,java,intellij+all,intellij+iml)would be a .gitignore template for Intell Idea, Eclipse, Java and Maven

Geolykt:
Eclipse user here - you should ALWAYS take a look at what you are staging before you commit, no matter your gitignore file.

lokka30:
Cheers for opening this issue - I'll commit a fix to this very soon. I've seen your PR although I've decided not to use the template directly, instead I've copied the file names that I think are relevant.

PR #18, opened 2021-10-10
Improve gitignore

MrNemo64:
As explained in the issue "[Improve .gitignore](https://github.com/lokka30/Treasury/issues/17)", the .gitignore file serves an important role in git development allowing for faster selection of files to commit, removing the need to exclude some files from every commit. However, the one in this repository is vague and only covers the Intell Idea IDE. With this gitignore we would cover Intell Idea, Eclipse, NetBeans, Maven and Java files that are probably not meant to be commited


lokka30:
Thanks for the PR. I'll close this since, using your template, I've made a more concise version that excludes unrelated plugins, IDEs and so on. Please let me know if my upcoming commit's gitignore modifications don't suit your intentions. :)

PR #19, opened 2021-10-10
Fix me.lokka30.treasury.api.economy.account.BankAccount#isBankOwner

MrNemo64:
The method `me.lokka30.treasury.api.economy.account.BankAccount#isBankOwner` was defined as
``` Java
default boolean isBankOwner(@NotNull UUID uuid) {
    return getOwningPlayerId() == uuid;
}
```
Since UUID is not a primary type comparing them with `==` will compare the memory address represented by the variable, this is however not what we want. For example,
``` Java
UUID owner = ... // owner of the accounr
BankAccount account = ...

// we have saved the owner into a database and now we retreat it
String ownerAsString = ...
UUID ownerFromString = UUID.fromString(ownerAsString);

assertTrue(account.isBankOwner(ownerFromString)); // fails
```
Instead we want to use the `equals` method to make sure we avoid these conflicts.

### DIFF
--- a/src/main/java/me/lokka30/treasury/api/economy/account/BankAccount.java
+++ b/src/main/java/me/lokka30/treasury/api/economy/account/BankAccount.java
@@ -48,7 +48,7 @@ public interface BankAccount extends Account {
      * @return whether the player with specified UUID owns the bank.
      */
     default boolean isBankOwner(@NotNull UUID uuid) {
-        return getOwningPlayerId() == uuid;
+        return getOwningPlayerId().equals(uuid);
     }
 
     /**


lokka30:
> use equals instead

Absolutely correct - thank you for noticing this!


I'd like to continue talking about this - currently I am stumped on a particular thought regarding bank owners and members coexisting. I've posted on SpigotMC, but please let me know if you would prefer to chat about it here instead :)

To avoid duplicating my messages across both sites I'll just add my thoughts on my Spigot post.

https://www.spigotmc.org/threads/new-economy-api-looking-for-your-feedback.527805/page-2#post-4279832

PR #26, opened 2021-10-11
Modify account events

MrNemo64:
With these changes we propose that AccountEvent should be a class itself and AccountTransactionEvent is a subclass of it, this way we can also create in the future AccountCreateEvent, AccountDeleteEvent, AccountAddNewMemberEvent, etc.  AccountEvent shouldn't implement Cancellable as in the future there can be an AccountEvent that can't be cancelled.
Another change proposed by this pull request is to override `getAccount()` in BankAccountTransactionEvent, NonPlayerAccountTransactionEvent and PlayerAccountTransactionEvent since these events only work with their respective account type it only makes sense that `getAccount()` returns their respective account types


lokka30:
Cheers @MrNemo64 - these changes are brilliant!

PR #37, opened 2021-10-17
Migrate to a Flow-like style

### LINKED ISSUE #36: Update the `/treasury migrate` command to use the latest concurrency code.
I have no experience *using* CompletableFutures yet. I also don't think the `migrate` subcommand has been seen by anyone, or at least commented on. So I think it would kill two birds with one stone if anybody could kindly assist me in updating `MigrateSubcommand.java` to use the new CompletableFuture methods (which your IDE should highlight as errors).

Cheers :)

### LINKED ISSUE #38: Address the difficulty of use with CompletableFutures, etc.
Please see #37 which is an in-progress resolution to this issue.

Jikoo:
Similar to using a `CompletableFuture` but ideally easier for users to parse and understand query results in the event of errors.

`EconomyProvider` and `Account` methods return an `EconomyPublisher` for their relevant types. The publisher itself does nothing until given an `EconomySubscription`. At that point the publisher will perform the action and provide the subscription the results. If the action needs a separate thread, the publisher should handle creating and running on it.

The subscription does not offer any way to block by design so that users won't perform blocking actions using it - they should perform any dependent operations in the subscription upon completion.

I did return to the use of an `Exception`, however, it's a generic that isn't actually thrown (depending on implementation of course) and doesn't fill in the trace. It's just a container for a `FailureReason`, message, and `@Nullable Throwable`. It could easily be removed in favor of a generic data container class. It's meant to handle things like SQL exceptions where Treasury really doesn't need an inbuilt reason for failure but people may want details.

For example, say I want to sell an Area to a player:
```Java
public void sellArea(@NotNull Player player, @NotNull Area area, double price) {
  UUID uuid = player.getUniqueID();

  econ.getPlayerAccount(uuid).subscribe(new EconomySubscription<PlayerAccount> {
    @Override
    public void accept(@NotNull PlayerAccount account) {
      // Pass bank account to helper method
      sellArea(player, account, area, price);
    }

    @Override
    public void error(@NotNull EconomyException exception) {
      // Tell player about error, halt execution.
      sendError(player, Errors.NO_BANK_ACCOUNT);
    }
  });
}

private void sellArea(@NotNull Player player, @NotNull PlayerAccount account, @NotNull Area area, double price) {
  account.withdrawBalance(price).subscribe(new EconomySubscription<Double> {
    @Override
    public void accept(@NotNull Double newBalance) {
      // Success! Player's money has been removed. Do the thing.
      area.setOwner(player.getUniqueID());
      sendMessage(player, Messages.PURCHASE_AREA, area.getName(), price, newBalance);
    }

    @Override
    public void error(@NotNull EconomyException exception) {
      // Tell player about error, halt execution.
      sendError(player, Errors.NOT_ENOUGH_MONEY);
    }
  });
}
```

It's not necessarily the most clean, however, it puts the obligation for handling threading on the implementation. Obviously error handling could actually use the detailed `EconomyResponse.Reason` (now `FailureReason` - successful operations pass non-null values to `EconomySubscription#accept`).

This is a WIP, but I feel like there's enough here to give an idea of what I'm talking about.
I partially updated the migration code but didn't finish since banks are even more work and the design isn't necessarily final (and maybe never will be acceptable). It's still a pretty big mess, fast and hacky. Currently player accounts should(TM) migrate, banks do not. Not generally a fan of commenting out swathes of code - that's what version control is for - but I wanted it to compile and not have to dig too deep for the old code in the history.

Closes #36 
Closes #38


lokka30:
This is awesome. I second Geolykt's opinion.

> Looking at the sample code as well as thinking how it can be implemented, I am a fan of this. It is relatively straightforward and actually forces developers to do error recovery.

Jikoo:
So a couple things on this that I wasn't sure about:

I copied the style of a [`Flow`](https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/Flow.html) `Publisher`/`Subscriber` for familiarity, however, this may not be the best approach. What happens if someone stores the `EconomyPublisher`? It may happily make the same request over and over again assuming the implementation allows it, but offering that as an option may not be desirable.
I think a better idea design-wise is to drop the return of a publisher entirely and instead just accept the `EconomySubscriber` as a parameter, returning void. It does promote a bit messier style, but it makes it harder to misuse. In this case I think I'd also rename `getX` methods `requestX` to make it clearer that you aren't immediately getting a result.

I've also done a bit of reading on `Exception` handling in my spare time. I think the `EconomyException` being an actual `Exception` is okay but unnecessary. The use of it in its current state is good performance-wise (not filling in stack, not in a try/catch), it just doesn't need to be one. I'm thinking about making it just a general `Failure` data holder.

Any input would be welcome.

Jikoo:
> Agree but we first need to close this pull request right? If not we are going to have conflicts.

I mean, I've already done one rebase on this in 06cb6b91e6e853e7f935297d63a318fa57e7e802, I don't really mind doing another if necessary. A relocation would be pretty simple to work around.

> So `treasury-api-<platform>` and` treasury-core` would depend from `treasury-api`; `treasury-core-<platform>` from `treasury-core` and `treasury-api-<platform>`; and `treasury-<platform>` from `treasury-core-<platform>` and would be the assembler? Or `treasury-<platform>` should have some code? And should they be childs from the same parent project or better to split them into projects and maybe move this repository into a new account/organization?

Any option works. For using JitPack separate repos may be easiest, though it is possible to use specific artifacts from JitPack.
If you're looking for examples of various styles, Vault uses separate repos (though only Bukkit has an implementation), WorldEdit uses one repo for all modules where each platform module builds itself, and OpenInv uses modules with an assembly module combining the final jar.
I think the best style guide for this particular case would likely be WorldEdit where the implementation for a particular platform builds itself. That should contain things like the JavaPlugin implementation, but being exclusively an API, Treasury generally does not fall prey to pitfalls that would make it difficult to migrate to a more modular system.

I feel like I've gotten a bit ahead of myself here though - I don't have any experience with Forge/Sponge, so I'm not sure if they even have a service system to benefit from Treasury availability. Clearly there's some degree of mod interoperability so it's possible, but would it be useful? Either way, probably a topic for another issue.

Jikoo:
> I'd like this to be a separate PR so I could analyse the difference between the two solutions and maybe evaluate for myself what the best one would be.

@Geolykt I've thrown together 4f320660c84cf8a1ac07f3ca9f02df6b4f0335ab for a quick and dirty preview.

MrIvanPlays:
@Jikoo you have a bunch of calls where calling map is unnecessary.

Jikoo:
> @Jikoo you have a bunch of calls where calling map is unnecessary.

We're not exactly at a code review point for a single quick and dirty commit I slapped together as fast as possible. That said, your point serves to highlight a hidden pitfall of a future style: ease of misuse.

MrIvanPlays:
> That said, your point serves to highlight a hidden pitfall of a future style: ease of misuse.

That's why documentation exists, support discords/ircs exist. Anyone who doesn't read the documentation and uses stuff by guessing what x does is probably gonna put his stuff on fire.

Jikoo:
> That's why documentation exists, support discords/ircs exist. Anyone who doesn't read the documentation and uses stuff by guessing what x does is probably gonna put his stuff on fire.

While users should read docs (which by the way don't exist there since I didn't bother writing them or thinking about logistics for more than a minute) there's a lot less potential for misunderstanding/misuse using the Flow style.

MrIvanPlays:
> there's a lot less potential for misunderstanding/misuse using the Flow style.

Sure but flow style dates waaaay back from java 7 ages where if you wanted something done async it needs a callback which looks ugly. See BungeeCord; it was written in that era, the ping response is like that, AsyncEvent is like that, etc., etc. You don't want to be a dinosaur, do you?

Jikoo:
> Sure but flow style dates waaaay back from java 7 ages where if you wanted something done async it needs a callback which looks ugly. See BungeeCord; it was written in that era, the ping response is like that, AsyncEvent is like that, etc., etc. You don't want to be a dinosaur, do you?

Actually, [Flow](https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/Flow.html) was introduced in Java 9. Prettiness is not the only factor to consider here.

MrIvanPlays:
> > Sure but flow style dates waaaay back from java 7 ages where if you wanted something done async it needs a callback which looks ugly. See BungeeCord; it was written in that era, the ping response is like that, AsyncEvent is like that, etc., etc. You don't want to be a dinosaur, do you?
> 
> Actually, [Flow](https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/Flow.html) was introduced in Java 9. Prettiness is not the only factor to consider here.

my bad, the way you implement flow is java 7 stylish.

Jikoo:
> my bad, the way you implement flow is java 7 stylish.

Expressiveness should not be confused with being dated. Calling anything that doesn't use tons of lambdas and chain calls similar to Java 7 is a pretty narrow view. Just because you can use a feature does not mean you always should. Take for instance local variable type inferencing from Java 10. It's certainly easier to type, but it can make the code much harder to follow without additional tools because you don't explicitly know what type of object you're dealing with. Tools do assist with this (as does using descriptive variable and method names), but the raw file readability decreases.

I agree that it is easier to write code fast using lambdas and chained calls. However, fast does not mean good (as can be evidenced by my own sample commit), particularly when it comes to encouraging beginners to properly implement the API.

Geolykt:
It should be noted that beginners will have a hard time grasping lambdas. I know that beginners are an overwhelming majority of bukkit, but it should be noted that they are also not a group to ignore. Catering to them may give the API some boost in credibility

MrIvanPlays:
> It should be noted that beginners will have a hard time grasping lambdas.

Cmon its not _that_ hard compared to kotlin's lambdas.
Also that you can do this 
```java
.handle(new Consumer<Double>() {

  @Override
  public void accept(Double value) {
    // do something
  }

});
```
lambdas in java in their current state are only for readability purposes, and that they look cool, nothing else. literally every 1 method interface (the method not being a `default` method, otherwise java will force you to expressively create a new instance) can be instantiated with a lambda.

Consider this example
```java
public interface ValueHandler {

  Something handle(Value value);
}
```
This can be fed into a method the following ways:
```java
// say we have a method somewhere that accepts ValueHandler

// can be fed like expressively creating a new instance
someMethod(new ValueHandler() {

  @Override
  public Something handle(Value value) {
    // ....
  }

});

// can be fed using lambda
someMethod(value -> {
  // ...
});

// can be fed using a separate class implementing ValueHandler
someMethod(new ValueHandlerImpl());
```
Lambdas can handle indefinite amount of arguments e.g
```java
// say we have a interface having 7 values which accepts
someMethod((value, type, alias, sender, whatever, something, blabla) -> {
  // ...
});
```

This is basic java everyone. And if a beginner bukkit/spigot developer doesn't know basic java, then how will he understand the bukkit/spigot api? You're throwing "beginner friendly" like this is a java library and not a bukkit/spigot one and that the developers who are going to use this won't know basic java.

Now back on topic. We've said the 2 pros of... wait actually let's create a pros & cons of futures and this implementation. Here are mines:

|                       | Pros                                                                                                                 | Cons                                                                                                            |
|-----------------------|----------------------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------|
| Current impl ("Flow") | declarative e.g  makes it clear that exceptions will happen and should be handled; cannot misuse methods that easily | Bulky (or no) chaining  (not really able to chain call and map to  other values with ease); overall messy usage |
| Futures               | chaining is easy; clearer and easy to maintain code                                                                  | ease of misuse       

A thing that both implementations have as a con is that if they're misused problems (especially exceptions) can be easily swallowed by the JVM (especially when those things are called asynchronously, our case).

I think this is the most exhaustive I can go. Now you decide. My opinion is and always will be: go with futures or future-like response objects.

MrNemo64:
> This is basic java everyone. And if a beginner bukkit/spigot developer doesn't know basic java, then how will he understand the bukkit/spigot api?

I disagree with this. The basics of java isn't the same as the basics of OOP. Basics of java are while, do-while and for loops, if-else statements, using lists, arrays, sets etc. The basics of OOP are abstracts classes, interfaces, etc. 
You dont _really_ need OOP to make a plugin that requires to use economy.
For a begginer lambdas are like magic. When I started with java I struggled a lot with lambdas. I saw that you declarated variables without saying its type and that made me so confused, like
```Java
something((i, element, th) -> { ... })
```
I couldn't understand this, what was `i`? What does it represent? Why doesn't have a type? At the time I didn't know that lambdas are an easy way to write interfaces that have only 1 no static no default method

MrIvanPlays:
> The basics of java isn't the same as the basics of OOP. 

This is where you're wrong. Java is an OOPL (**O**bject **O**riented **P**rogramming **L**anguage). Except knowing what you described as "java basics" you have to know what an object is, what object types java has and how they works. Otherwise you don't know basic java. 
Not going to talk about abstraction. Abstraction is a whole another topic and I am not considering it basic java. 
About seeing the object type of an variable in a lambda, if you are using the good ide (IntelliJ), you can Ctrl + Click on it and see what the fed object is. 

> For a begginer lambdas are like magic. 

No one forces you to use lambda. As I explained, you can feed the method by expressively creating an instance of the needed interface.

MrNemo64:
> This is where you're wrong. Java is an OOPL (**O**bject **O**riented **P**rogramming **L**anguage). Except knowing what you described as "java basics" you have to know what an object is, what object types java has and how they work. Otherwise you don't know basic java.

Knowing what an object is and how they work on Java makes developing easyer but you don't necesarily have to know it. The basics of java doesn't include lambdas 

I'd like to add to the pros-cons table that Futures can use more resorces, you need to store the map funcion, default value, exception handler and since you also have to check if those are null. Realisticly speaking, this won't have any impact at all, but I wanted to mention it.

Personally I like both options and would be happy with both

MrIvanPlays:
> but you don't necesarily have to know it.

This applies for languages which are not object oriented e.g C, Python, JavaScript, etc. But this is not Python. This is Java. Again, Java is an OOPL. If you don't know objects then how would you do a program? In languages such as python you can do stuff like this:
```python
print("Hello, World!")
```
and it will just work out of the box. But in Java, you have to wrap that into an object, then into a method and use a desired object to get another object to do the heavylifting about doing something e.g
```java
public class MyFirstJavaProgram {

  public static void main(String[] args) {
    System.out.println("Hello, World!");
  }

}
```
You have to know it and understand it. This is a basic thing of java - the objects. Because, again, it is an OOPL.

> Futures can use more resorces

Right after you're done with it the GC picks it up. Except of course if you are not using the proper GC (that being G1). Such objects in memory are kilobytes, and they will do an impact if they are tens of thousands or more at once, which I'm 100% sure it won't happen.

> Personally I like both options and would be happy with both

There's no middle ground in programming. Either this or that. So make your choice.

Geolykt:
> > Personally I like both options and would be happy with both
> 
> There's no middle ground in programming. Either this or that. So make your choice.

Actually yes, in theory they should do the same thing so you should be able to write a wrapper that delegates the calls from one way of using it to the other

lokka30:
> 
> 
> > > Personally I like both options and would be happy with both
> > 
> > 
> > There's no middle ground in programming. Either this or that. So make your choice.
> 
> Actually you can do both, in theory they should do the same thing so you should be able to write a wrapper that delegates the calls from one way of using it to the other

If we could have such a wrapper then we would have the best of both worlds. I sure hope this is possible since I would love to use Jikoo's flow implementation.

MrIvanPlays:
So you say that this is "flow" style. Well, @Jikoo, you killed the baby you created by yourself. The javadocs link mentioned some replies ago pushes me to the fact that this is not "flow" style but a java 7 way of implementing a callback. I've utilized the `Flow` class in java 17 and this is a test I created: https://srcb.in/VdxsM9mzK4

So,
d4d-flow

Jikoo:
> So you say that this is "flow" style. Well, @Jikoo, you killed the baby you created by yourself. The javadocs link mentioned some replies ago pushes me to the fact that this is not "flow" style but a java 7 way of implementing a callback. I've utilized the `Flow` class in java 17 and this is a test I created: https://srcb.in/VdxsM9mzK4
> 
> So, d4d-flow

Are you really not able to see the parallels between [Flow.Subscriber](https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/Flow.Subscriber.html) and the EconomySubscriber? There used to be a publisher, but if you bothered to read the PR's history you would have encountered https://github.com/lokka30/Treasury/pull/37#issuecomment-946904337 containing my reasoning for removing it. The EconomySubscriber is a single element subscription to be completed at a later date.

> you killed the baby you created

You do realize this wasn't what I wanted in the first place, right? I wanted Treasury to just return `CompletableFutures`. This whole PR was initially just an example of a potential compromise to make it easier for people to use.

I do find it amusing that you've selected someone famous for taking controversial stances and then turning tail to run whenever anyone actually engages him in a debate for your image of choice. Quite fitting given your only defense is shouting "Java 7" or "dinosaur" if anyone disagrees with you.
Lashing out at people doesn't reinforce your point, it just makes you look unprofessional and unreliable.

> If we could have such a wrapper then we would have the best of both worlds. I sure hope this is possible since I would love to use Jikoo's flow implementation.

It would definitely be possible to write a wrapper that uses `Futures` (or a more specific interface if that's desirable). I can work on that next.

MrIvanPlays:
I'm sorry for hurting your feelings but this is how it stands.

> Quite fitting given your only defense is shouting "Java 7" or "dinosaur" if anyone disagrees with you.

Someone didn't read my big post from start to end.

> you've selected someone famous
> Lashing out at people doesn't reinforce your point

I'm just giving examples.

> it just makes you look unprofessional and unreliable.

HEY! LETS REMOVE HIM FROM EXISTING BECAUSE OH MY GOD HE HAS AN OPINION!!!!!!
On a more serious note, I'm giving my opinion and throwing some examples and facts here and there to say where I'm basing my opinion from. I am not forcing anyone to agree with it.

> It would definitely be possible to write a wrapper that uses `Futures` (or a more specific interface if that's desirable). I can work on that next.

Would be much easier and cleaner to do `Future` -> `callback` .

Jikoo:
> Someone didn't read my big post from start to end.

I did. I thought it was pretty valid overall but it lacks perspective. Think about Spigot Plugin Development - how many of those questions are from advanced users? Maybe 1/20, and that's being generous. That's the target market for this API. There are a lot of very knowledgeable people there, but there are also an overwhelming number of people who are writing their very first plugin and have no idea what Java even is. Is an economy plugin a bad first project? Probably, but hey, everyone starts somewhere. My initial embarrassments in plugin development are fully on display if you scroll down my repos list.

> I'm just giving examples.
>
> HEY! LETS REMOVE HIM FROM EXISTING BECAUSE OH MY GOD HE HAS AN OPINION!!!!!! On a more serious note, I'm giving my opinion and throwing some examples and facts here and there to say where I'm basing my opinion from. I am not forcing anyone to agree with it.

I appreciate your input (and agree with you in a lot of cases!), however, your phrasing is often overly aggressive (see in particular the Java version issue) and makes it clear that you haven't followed the entirety of the evolution of the PR. Passion about something is well and good, but I've certainly elected to not use software before because of the devs' behavior. A modicum of professionalism is important even if you're not trying to find a career in the field.

> Would be much easier and cleaner to do `Future` -> `callback` .

I agree. However, I believe that the primary API, the one not wrapped, should be most simplistic one to make life easier for users. You've pointed out yourself that an extra `Future` is a somewhat negligible amount of memory, so making the `Future `the wrapper instead is hardly a breaking issue.

MrIvanPlays:
> makes it clear that you haven't followed the entirety of the evolution of the PR.

not trying to hide this. I haven't been here since the very start and I apologize for parts of my replies about flow style. I wasn't aware that the first stages of this PR implemented it the way it should, but then got redone because of whatever.

> but I've certainly elected to not use software before because of the devs' behavior.

Quick example: you know md_5 very well, he is someone who doesn't like improving stuff, just see the state of BungeeCord and Spigot (hence why Paper and Waterfall exist). He likes control; an example about that: spigot is an open source project, right? well he's implementing snapshot stuff privately and then shouts "our amazing dev team (realistically only him) updated spigot to 1.<insert version here>". that's like saying you did it as a whole and that no one helped you with this, but the truth is that you don't give the opportunity to the people to help.

So now that you've understand what a person behind md_5 is, why do you continue using spigot, spigot based fork(s) and BungeeCord? <font size="1">I know why and that's because they're here from ages and have tons of plugins and information other software doesn't, say Velocity, which in fact is the replacement of BungeeCord and Waterfall if you want much better support and performance.</font>

You shouldn't limit yourself to what to use and what not to use. Does the software has a clear documentation? yes? does it do the job you want? yes? then why do you want to code this by yourself when it has been already coded?

> I believe that the primary API, the one not wrapped, should be most simplistic one to make life easier for users.

Sure but think about how ridiculous a flow -> future wrapper when used would look like?

```java
FutureWrapper.wrap(economy.setBalance(0.0d, new EconomySubscriber<Double>() { // ... })
  .<some future methods>
```

is this how I will have to use this API?

Jikoo:
> > makes it clear that you haven't followed the entirety of the evolution of the PR.
> 
> not trying to hide this. I haven't been here since the very start and I apologize for parts of my replies about flow style. I wasn't aware that the first stages of this PR implemented it the way it should, but then got redone because of whatever.

Well, I apologize for my own unprofessionalism, I certainly could have reacted more graciously.

> Quick example: you know md_5 very well, he is someone who doesn't like improving stuff, just see the state of BungeeCord and Spigot (hence why Paper and Waterfall exist). He likes control; an example about that: spigot is an open source project, right? well he's implementing snapshot stuff privately and then shouts "our amazing dev team (realistically only him) updated spigot to 1.". that's like saying you did it as a whole and that no one helped you with this, but the truth is that you don't give the opportunity to the people to help.
> 
> So now that you've understand what a person behind md_5 is, why do you continue using spigot, spigot based fork(s) and BungeeCord?

I don't personally use Spigot. I don't run servers any more (I retired from public server administration in 2017) unless my core friend group wants a private server for a MC update, but if I did I'd use at least Paper, preferably Purpur.
Spigot is kind of terrible - look at the fact that Bungee's chat component API *still* isn't included in the Bukkit API (`Player.Spigot` etc. hasn't made any sense since Spigot took over for Bukkit) and there's no API-safe way to create an item hover even using the Spigot API instead. It's stagnant, people are absolutely discouraged from attempting to contribute.

The only reason I still program against Spigot-API is that Spigot has too big a market share of Bukkit implementations for me to ignore it. While I don't post most of my resources to Spigot/Bukkit/Whatever other hosts, I do want them to be accessible if someone does happen to stumble on them and think they're useful. If Paper ends up hitting 90% of servers I'll probably drop Spigot support.

Honestly overall one of the reasons you don't see me in any of the live chat servers (IRC, Discords, whatnot else) is that I find the community in general to be too drama-filled for my taste. Paper has its own internal drama separate from the Spigot vs Paper fight (which is hilarious considering the original Bukkit vs Spigot fight, history repeating itself). I'm just here to have fun programming in my free time, I don't need it.

> You shouldn't limit yourself to what to use and what not to use. Does the software has a clear documentation? yes? does it do the job you want? yes? then why do you want to code this by yourself when it has been already coded?

In the cases where I've given software a miss there's always been an at least semi-viable alternative. Maybe not quite as good, but much more personally satisfying. If I wasn't already invested in Bukkit plugin development I'd probably have jumped ship to Sponge or writing exclusively server-side Forge mods a long time ago.

> Sure but think about how ridiculous a flow -> future wrapper when used would look like?
> 
> ```java
> FutureWrapper.wrap(economy.setBalance(0.0d, new EconomySubscriber<Double>() { // ... })
>   .<some future methods>
> ```
> 
> is this how I will have to use this API?

That sounds pretty gross. I was thinking more along the lines of a full `EconomyProvider` wrapper.

```Java
WrappedEconomy economy = new WrappedEconomy(economyProvider);

economy.setBalance(0.0d)
    .// some future methods
```
So the implementation would look like this:
```Java
public class WrappedEconomy {

  private final EconomyProvider economy;

  public WrappedEconomy(EconomyProvider economy) {
    this.economy = economy;
  }

  public FutureImpl<Double> setBalance(double balance) {
    FutureImpl<Double> future = new FutureImpl<>();
    economy.setBalance(balance, new EconomySubscriber<> {
      // success -> future success
      // fail -> future fail
    });
    return future;
  }

}
```

However, when you put it that way, it could actually be done with a `Consumer<EconomySubscriber<T>>`...
```Java
FutureWrapper.wrap(subscriber -> economy.setBalance(0.0d, subscriber))
    .// some future methods
```

Either way the wrapper would definitely always handle the creation of the subscriber, having the `Future` user define it defeats the purpose of making it a future.

MrIvanPlays:
> However, when you put it that way, it could actually be done with a `Consumer<EconomySubscriber<T>>`...
> 
> ```java
> FutureWrapper.wrap(subscriber -> economy.setBalance(0.0d, subscriber))
>     .// some future methods
> ```

That is a much cleaner way of wrapping it.

MrNemo64:
Couldn't we provide an implementation of the subscriber that you can use as a future? On the complete method calls the map function and all of that and on the error method calls the error consumer

MrIvanPlays:
> Couldn't we provide an implementation of the subscriber that you can use as a future?

No. The return value of the method must be a future. Otherwise it will look as gross as just using flow style.

MrNemo64:
> > Couldn't we provide an implementation of the subscriber that you can use as a future?
> 
> No. The return value of the method must be a future. Otherwise it will look as gross as just using flow style.

```Java
requestX(uuid, new EconomySubscriberWarper().
                          .defaultValue(...)
                          .mapTo(...)
                          .ifError(...));
```

```Java
EconomySubscriber es = new EconomySubscriberWarper().
                          .defaultValue(...)
                          .mapTo(...)
                          .ifError(...));
requestX(uuid, es);
```
Is not that bad, kinda like a builder. 

Maybe we could have 2 request methods. 1 returns the future like object and the other takes a economy subscriber and one of them is just a default methods that uses the other. Maybe something like
```Java
FutureEconomyResponse<X> requestX(uuid); 

default void requestX(uuid, EconomySubscriber<X> subscriber) {
  requestX(uuid).then(subscriber::accept).exceptionally(subscriber::onError);
}
```
And  the implementation would do
```Java
FutureEconomyResponse<X> requestX(uuid) {
  return FutureEconomyResponse.supplyAsync(() -> { ... });
  // maybe even have a FutureEconomyResponse.supply that runs sync for when the value is already cached?
}
```
Although it may be confusing as to why there are 2 methods with the same name

MrNemo64:
> 1 method is better than 2

Agree but, anyone has another idea?

lokka30:
I think we should only have one method that is used to get the result.

MrNemo64:
> I think we should only have one method that is used to get the result.

I also think so and using the EconomySubscriberWarper seems like the best option to me

MrIvanPlays:
>>> I am now seeing a potential source of confusion with the idea to move from "get" to "request" to clarify usage for users - if I request a player account, it could be interpreted as requesting a new player account, not only an existing one (even though the docs do specify). Does anyone have any ideas on that front? Any more expressive words that identify a future value return as an operation that does not immediately return a value without being excessively verbose? Should we revert to "get" instead?


retrieve

MrIvanPlays (approved):

NoahvdAa (approved):

src/main/java/me/lokka30/treasury/api/economy/response/FailureReason.java
```diff
@@ -0,0 +1,138 @@
...
+     * @see me.lokka30.treasury.api.economy.account.Account#depositBalance(double, me.lokka30.treasury.api.economy.currency.Currency, EconomySubscriber)
+     * @since v1.0.0
+     */
+    NEGATIVE_AMOUNT_SPECIFIED,
+
+    /* Currencies */
+
+    /**
+     * A constant representing failure due to the inability to locate an
+     * {@link me.lokka30.treasury.api.economy.currency.Currency Currency}.
```
MrNemo64, line 121: Shouldn't this be "A constant representing failure due to the inability to locate **a** {@link me.lokka30.treasury.api.economy.currency.Currency Currency}."

lokka30: It should - thanks for the find.

src/main/java/me/lokka30/treasury/plugin/command/treasury/subcommand/migrate/AccountMigrator.java
```diff
@@ -0,0 +1,83 @@
...
+
+    default void migrate(
+            @NotNull Phaser phaser,
+            @NotNull T fromAccount,
+            @NotNull T toAccount,
+            @NotNull MigrationData migration) {
+        for (Map.Entry<Currency, Currency> fromToCurrency : migration.migratedCurrencies().entrySet()) {
+            Currency fromCurrency = fromToCurrency.getKey();
+            Currency toCurrency = fromToCurrency.getValue();
+            CompletableFuture<Double> balanceFuture = new CompletableFuture<>();
```
MrIvanPlays, line 42: think this logic can be migrated to EconomySubscriber.asFuture

MrNemo64 (approved):

src/main/java/me/lokka30/treasury/plugin/command/treasury/subcommand/migrate/AccountMigrator.java
Jikoo: It could be, though not directly. I'm reluctant to do that without doing a full pass over the entirety of migration, which I don't really have time to do in the coming week - life got busy, unfortunately.

MrIvanPlays: gotcha.

MrNemo64: I don't see why need this. Since what we have works now and Jikoo is busy we may as well do it in another pr so we can close this one and continue working on other things.

lokka30 (approved):

lokka30:
I am ready to merge this whenever it's ready. Even if there are missing parts, we can always add them in after it's merged. So long the core concepts are done. Thank you to everyone for your contributions and input!

Jikoo:
Should be set on my end, I believe I resolved all change requests.

MrNemo64 (approved):

Issue #27, opened 2021-10-11
Fix JavaDoc formatting


MrNemo64:
Currently the Javadoc format follows a pattern: author - see - since - description - parameters - throws - return
This format creates the problem that JavaDoc genertors think that the description is part of the see tag, we can see this by generating the javadoc.
![Image of generated JavaDoc](https://i.imgur.com/0xPQV9I.png)
Another problem with it is using formatting options that don't exists in the JavaDoc like \`Marked` This JavaDoc was generated using the java 8 jdk

lokka30:
> 
> 
> Currently the Javadoc format follows a pattern: author - see - since - description - parameters - throws - return This format creates the problem that JavaDoc genertors think that the description is part of the see tag, we can see this by generating the javadoc. ![Image of generated JavaDoc](https://camo.githubusercontent.com/56bf1609991210a7ac52a7119edaae39493f01e1ad15c84e99f01456c5532def/68747470733a2f2f692e696d6775722e636f6d2f307850515639492e706e67) Another problem with it is using formatting options that don't exists in the JavaDoc like `Marked` This JavaDoc was generated using the java 8 jdk

Thanks for bringing this to my attention. I'll for sure fix the javadoc formatting throughout the resource.

However, the `Marked` formatting is intentional - even if they are not formatted in the JavaDoc I still find these characters useful to highlight a specific part of texts i.e. class names.

( This issue is linked to #4. )

Geolykt:
In this case, use `{@link ClassName}`

lokka30:
> 
> 
> Currently the Javadoc format follows a pattern: author - see - since - description - parameters - throws - return This format creates the problem that JavaDoc genertors think that the description is part of the see tag, we can see this by generating the javadoc. ![Image of generated JavaDoc](https://camo.githubusercontent.com/56bf1609991210a7ac52a7119edaae39493f01e1ad15c84e99f01456c5532def/68747470733a2f2f692e696d6775722e636f6d2f307850515639492e706e67) Another problem with it is using formatting options that don't exists in the JavaDoc like `Marked` This JavaDoc was generated using the java 8 jdk

To solve this issue, do my comments go above all of the Javadoc tags?

MrNemo64:
Since we now have a EconomyAPIVersion enum maybe the \@since tag could have a link to the enum entry corresponding to the version. So for example for something added in v1.0.0
```Java
/**
 * ...
[...]

Issue #14, opened 2021-10-05
Work on a Chat API


lokka30:
I'd like to implement a Chat API into Treasury.

I'm very interested in what anyone has to say about this, since this is the perfect opportunity to create a great Chat API.

I'm unsure when this will be added, and the Economy API is heavily prioritized over this.

lokka30:
Note: the `developer thoughts wanted` label requests any ideas from developers regarding this API. What should people expect from a new Chat API? What do existing Chat APIs do incorrectly?

MrNemo64:
I've been thinking about this and this is what I came with:
1. Maybe making this a provider-consumer system isn't a great idea, how many chat providers are there?
2. Multi lang support: I would like to be able to tell to the chat API where my lang files are and the API will manage them, for example:
```Java
TreasuryChat chatAPI = ...
Plugin myPlugin = ...
File myLangFolder = ... // where all my lang files are
String defaultLangFile = "en_US.yml" // default lang file to use
chatAPI.registerLangFolder(myPlugin , myLangFolder, defaultLangFile );
```
Now that the API knows where my lang files are I would be able to do the following:
```Java
TreasuryChat chatAPI = ...
Player player = ...
Plugin myPlugin = ...
Map<String, String> variables = ... // variables of the message
chatAPI.sendMessage(myPlugin, player, "messages.welcome", variables);
```
Now the chat API will do the following:
  1. Get the language of the player (maybe a player can specify its language with a command or maybe using [`Player#getLocate()`](https://hub.spigotmc.org/javadocs/spigot/org/bukkit/entity/Player.html#getLocale()))
  2. The chat API will check if there is a lang file in my lang folder with the same language as the players
    - If the lang file is found and it has the message, uses that message
    - If the lang file isn't found/doesn't have the entry, the default lang file is used
    - If the message is still not found throw a RuntimeException
  3. The chat API colors, replaces variables, etc and sends the message
 
I've seen that some chat plugins allow for chat channels so you join a channel and only recieve messages sent on that channel. Don't know if this is worth doing.
Maybe the API should allow to make messages with click and hover events
Maybe a mails/memory system should be done: being able to send a message to a player even if its offline. When the player joins all those messages are sent/they are told that they have messages to read (kinda like essentials mail system)

This is what I thought so far

lokka30:
[...]

PR #45, opened 2021-11-12
Split Treasury's code into modules

### LINKED ISSUE #42: Split Treasury's code into modules
I believe this was discussed in #37.
This issue will be on hold until #37 is completed.

I haven't used modules before, I might need pointers here. I'm also very interested to hear what modules developers think Treasury should have. I'm thinking:

* `treasury-api`
* `treasury-plugin-bukkit`, `treasury-plugin-sponge`, etc

MrIvanPlays:
Intends to close #42


api/src/main/java/me/lokka30/treasury/api/economy/misc/APIUtils.java
```diff
@@ -0,0 +1,19 @@
...
+    /**
+     * @author lokka30
+     * @since v1.0.0
+     * If the specified amount is less than zero
+     * then zero is returned. Otherwise, the amount
+     * is returned. This ensures an amount is at least zero
+     * since negative values are not allowed in the API.
+     * @param amount to check for.
+     * @return the unmodified or modified amount.
+     */
```
MrNemo64, line 15: This javadoc is icorrect. The correct order is first description then tags. If not the javadoc parser thinks that the description is, in this case, inside the since tag

MrIvanPlays: 1. This is going to be long until it is completed
2. This has been copied like that.

MrNemo64: Yeah I've been checking other files and most of them have this problem. I've also though that maybe the \@since tag should have a link to the EconomyAPIVersion corresponding to it, I'll mention this on #27 to see what others think

MrIvanPlays: there will be a separate PR to fix all javadoc problems.

MrNemo64:
I've checked the code and since the migration logic is the same, it only changes the trigger that starts it, for example the trigger for bukkit/spigot/paper is a bukkit command but for other servers with their own command system the trigger would be their own command; the migration logic could be in the core module.
I'd also rename the Bukkit module into Bukkit-core so you know it has the bukkit plugin logic

MrIvanPlays:
As I said, this is not a completed pull request. Stuff which can be abstracted are yet to be abstracted. Some general API stuff will probably need to be changed. It is not complete so please do not do any reviews, only suggest stuff to be abstracted.

Geolykt:
I think that it might be better to split the modules over multiple repos, but that may also be annoying for others

MrNemo64:
> I think that it might be better to split the modules over multiple repos, but that may also be annoying for others

Most repos I see have them all in one place being childs of a common parent. 
The problem with having them together is that for the simplest of changes you need to clone everything, that can be anoying
Having them separated doesnt seem like much of a issue to me

MrIvanPlays:
No. The reason why they're called modules is that you contain them at 1 place on the internet.

Geolykt:
Then we will run into issues with forge + fabric should we ever choose to support them. It takes ages to compile those

MrIvanPlays:
Why should we support fabric/forge? Either way they're hassle and if we so so we'd have to switch to using gradle

Geolykt:
> Why should we support fabric/forge? Either way they're hassle and if we do so we'd have to switch to using gradle

Because we fully can. I also believe that having a cross-plattform economy API for once would help immensely.

MrIvanPlays:
In that case we'd need to migrate to gradle, which is a hassle for multiple modules and lots of other things. But we need @lokka30 on that. Fairly sure we don't need fabric/forge support and gradle.

Geolykt:
Yeah, right now it is not that good to have that support; was just throwing the idea around

MrNemo64:
I haven't really seen mods with an economy system thing or that needs an economy provider, they all normally just add the items that they need and not depend on anything else, and if supporting them means switching to Gradle it may be a little to much for so little, I can't think of mods that need an economy interface but I'm not really into mods. It may make more sense to support servers that allow the usage of mods and plugins

MrIvanPlays:
Mods are to add stuff which vanilla minecraft can't offer or to optimize minecraft's internals. They're not used for such simple additions.

lokka30:
- I don't believe Treasury will have mod support in the future. Happy to include it if we ever run out of things to do, though. :smile:
- I'm against splitting Treasury into multiple repositories. How come would we need to do this? Or is this only for the Forge/Fabric support?
- I very briefly gave an attempt at using Gradle once and it failed miserably. If there are decent benefits with Gradle then we can make the switch. I know it does compile much faster.

MrNemo64:
> * I'm against splitting Treasury into multiple repositories. How come would we need to do this? Or is this only for the Forge/Fabric support?

This is not necesarily only for forge/fabric support. Instead of a parent project and multiple modules we would just have multiple modules that depend on each other



> * I very briefly gave an attempt at using Gradle once and it failed miserably. If there are decent benefits with Gradle then we can make the switch. I know it does compile much faster.

The only benefit I see with gradle instead of maven is compile time, but lets be real does it take that much time to compile anyways? If I'm not mistaken Gradle has more bugs, less repositoryes and there are easy ways to use a maven repo on a gradle project but not otherways (may be wrong here). Gradle also tends to be harder to use

lokka30:
> This is not necesarily only for forge/fabric support. Instead of a parent project and multiple modules we would just have multiple modules that depend on each other

Wouldn't the API be a parent module?

> The only benefit I see with gradle instead of maven is compile time, but lets be real does it take that much time to compile anyways? If I'm not mistaken Gradle has more bugs, less repositoryes and there are easy ways to use a maven repo on a gradle project but not otherways (may be wrong here). Gradle also tends to be harder to use

Compile time is a very minor developer luxury - although Maven takes a few seconds longer than I think it should, it's really far from enough to justify moving to Gradle. I assume more people know how to use Maven than Gradle as well.

MrIvanPlays:
Gradle doesn't have their own repository schema. They use maven's repositories. Benefits of gradle are compile time and that you can make custom actions when compiling - say to validate something (maven can also validate stuff via the plugins, but is kinda limited, so dont make it a solid plus). Also gradle has a lot of plugins and you can make your own plugins for it. Unfortunately, gradle is kind of a hassle when it comes to deploying to a maven repo (it can, but it is buggy), having multiple modules with it is also not really a go-to. We're sticking to maven as it is easy to use.

Geolykt:
From my experience gradle has been far easier for deployment than maven as latter does not create the checksums. But I am also using a highly unusual publishing system where I push to mavenLocal first and then manually copy them over to the webserver

MrIvanPlays:
I think we've decided about what build system we will use. End of discussion maven vs gradle.

Now you may want to check out the newest changes and suggest things on the put `// todo`'s where help is needed.

MrNemo64:
> I think we've decided about what build system we will use. End of discussion maven vs gradle.
> 
> Now you may want to check out the newest changes and suggest things on the put `// todo`'s where help is needed.

Can you explain a little your thought process?

MrIvanPlays:
I'm aiming to abstract as much as it can be abstracted into the core module so it is easier to implement treasury on different platforms. 

There's a lot to do in here ; I plan to document every part of that core module so it is 
1. easy to maintain
2. easy on implementing different platforms
3. easy on someone who wants to do a simple fix

Plan is the following:
1. Get rid of microlib (sorry, lokka30, but microlib's too bukkit oriented), which means to implement : 
 - platform independent messaging system
 - platform independent settings handling
 - platform independent update checker
2. Implement platform independent getter for economy provider and provider registrar data.
3. Implement platform independent command system.
4. More stuff to do on the code (can't think of more stuff as of right now)
5. Work on maven : 
  - deployment profiles for codemc and repo.mrivanplays.com
  - compile in the jar only what we need, provide only otherwise.

MrNemo64:
> * platform independent messaging system
> * platform independent settings handling
> * platform independent update checker

Do we really need to implement them? The bukkit version of the plugin will check on spigotmc but the paper one may check on papermc and so on. Update checker probably can be on the core module of each platform
The settings handler and messaging system are the ones that seem like they need an abstracted system but mainly so we can have the migration login on the core module.



> 1. Implement platform independent getter for economy provider and provider registrar data.

I dont understand what you mean here. So in bukkit we have the service provider that would allow to get the economy provider but in other platforms there may not be one so we give the responsability of providing the provider to a new provider?

It may make sense to, before doing mayor changes to the code, discuss what needs to be on what module and what do we need to abstract

MrIvanPlays:
> Do we really need to implement them? 

Yes. If we do it per platform then we will have lots of duplicated code, which is something you don't want. 

> So in bukkit we have the service provider that would allow to get the economy provider but in other platforms there may not be one so we give the responsability of providing the provider to a new provider?

No. See the commands and the other parts of the code use the `EconomyProvider` class. We need an instance of it. That's what this means.

lokka30:
> Get rid of microlib (sorry, lokka30, but microlib's too bukkit oriented), which means to implement :

If we add more platforms for Treasury, go ahead, I really don't mind :)

> service provider

What if we made our own service provider class, so it can be used universally?

MrIvanPlays:
I mean, not that we can't, the thing is that we don't really need it. Sure, we'd remove that provider from the core module, but at the cost of maintaining a singleton API.

MrNemo64:
Im seeing that most of the changes right now are message related. It may make sense to also consider #14 while making this changes

MrIvanPlays:
> Im seeing that most of the changes right now are message related. It may make sense to also consider #14 while making this changes

That's about holding prefixes/suffixes, nothing to do with what this messaging system is. This messaging system is about treasury plugin's messages. What I'm doing is not and should not be part of the API.

MrIvanPlays:
Code's now basically ready for a review - the only things that need fixing are the way config versions are handled - which is certainly not the way to go.

Maven have not been done yet - this is the next part of the implementation.

MrIvanPlays:
What do you think about migrating messages and config generation to [AnnotationConfig](https://github.com/MrIvanPlays/AnnotationConfig) ? That way we're abstracting 1 more layer from platform implementation, and it also automatically handles missing entries and, if wanted, whole config generation.

lokka30:
> 
> 
> What do you think about migrating messages and config generation to [AnnotationConfig](https://github.com/MrIvanPlays/AnnotationConfig) ? That way we're abstracting 1 more layer from platform implementation, and it also automatically handles missing entries and, if wanted, whole config generation.

It looks promising, but I am unsure. How does it handle existing comments when it saves to files - are the comments kept or removed?

MrIvanPlays:
> How does it handle existing comments when it saves to files - are the comments kept or removed?

kept. https://tryitands.ee/ if you don't believe me.

MrIvanPlays:
I've tested the plugin and fixed obvious errors, and the last things to do:

- `common.states.yes` and `common.states.no` messages don't load for some weird reason
- CodeMC deployment profile
- missing options generation

The only thing I didn't test fully was the migration subcommand, because it needs economy providers I cba to make myself.

lokka30:
Thank you very much. In a week or so I will have all of the free time I could ask for, I'll be able to take a proper review then. 🍻

MrIvanPlays:
The problem for common.states found: bukkit parses `yes` and `no` to `true` and `false` . This is probably a bug with bukkit but i cba to report it, even if i do so md_5 will just "but what about old plugins and bukkit being forward support"

I'm renaming these keys to `common.states.does` and `common.states.does-not`.

lokka30:
> 
> 
> The problem for common.states found: bukkit parses `yes` and `no` to `true` and `false` . This is probably a bug with bukkit but i cba to report it, even if i do so md_5 will just "but what about old plugins and bukkit being forward support"
> 
> I'm renaming these keys to `common.states.does` and `common.states.does-not`.

Ooh, good find. I think I recall seeing that issue once, unsure if it something with SnakeYAML or even the YAML specification itself.
I'm unsure about the replacements but we can change that later if there is a possible improvement.

lokka30:
@MrNemo64 @Geolykt @Jikoo @NoahvdAa - if you are interested: a review on this PR of however much depth you would like would be appreciated. I haven't been able to check out the changes properly yet, I'll have all the free time in the world in about a week to pick Treasury back up again.

MrIvanPlays:
> > The problem for common.states found: bukkit parses `yes` and `no` to `true` and `false` . This is probably a bug with bukkit but i cba to report it, even if i do so md_5 will just "but what about old plugins and bukkit being forward support"
> > I'm renaming these keys to `common.states.does` and `common.states.does-not`.
> 
> Ooh, good find. I think I recall seeing that issue once, unsure if it something with SnakeYAML or even the YAML specification itself. I'm unsure about the replacements but we can change that later if there is a possible improvement.

Well it's changing the keys so probably SnakeYAML.

> a review on this PR of however much depth you would like would be appreciated.

I agree a review is needed

bukkit/src/main/java/me/lokka30/treasury/plugin/bukkit/BukkitTreasuryPlugin.java
```diff
@@ -0,0 +1,255 @@
...
+
+    @Override
+    public @NotNull Settings getSettings() {
+        return this;
+    }
+
+    public String colorize(@NotNull String message) {
+        return fork.isSpigot()
+                ? net.md_5.bungee.api.ChatColor.translateAlternateColorCodes('&', message)
+                : org.bukkit.ChatColor.translateAlternateColorCodes('&', message);
```
NoahvdAa, line 185: Is there a real difference between these two?

~~also, ew, legacy color codes~~

bukkit/src/main/java/me/lokka30/treasury/plugin/bukkit/MessagesImpl.java
```diff
@@ -0,0 +1,56 @@
...
+                return MessagesConfigAccessor.EMPTY;
+            }
+            YamlConfiguration config = YamlConfiguration.loadConfiguration(file);
+            return (MessagesConfigAccessor) config::get;
+        };
+    }
+
+    @Override
+    public void generateMissingEntries(@NotNull Collection<MessageKey> keys) {
+        // todo: find another way!!!
```
NoahvdAa, line 45: [Configurate](https://github.com/SpongePowered/Configurate) can do this, with comment support :)

EDIT: Am I seeing this wrong or does your current implementation just delete all user-entered messages if the missing entries are generated?

bukkit/src/main/java/me/lokka30/treasury/plugin/bukkit/command/TreasuryCommand.java
```diff
@@ -0,0 +1,60 @@
...
+        this.sources = sources;
+    }
+
+    @Override
+    public boolean onCommand(
+            @NotNull CommandSender sender,
+            @NotNull Command command,
+            @NotNull String label,
+            @NotNull String[] args
+    ) {
```
NoahvdAa, line 38: What the hell happened with this formatting

core/src/main/java/me/lokka30/treasury/plugin/core/config/messaging/MessageKey.java
```diff
@@ -0,0 +1,272 @@
...
+                            "%prefix% Invalid usage, try ''&b/%label% migrate <providerFrom> <providerTo>&7''.",
+                            "%prefix% Valid economy providers: &b%providers%&7."
+                    )
+            ),
+            ConfigVersion.V1
+    ),
+    MIGRATE_REQUIRES_TWO_PROVIDERS(
+            "commands.treasury.subcommands.migrate.requires-two-providers",
+            new Messages.MessageHolder(
+                    "%prefix% You can''t use this subcommand unless you have 2 economy providers running."
```
NoahvdAa, line 151:
```suggestion
                    "%prefix% You can''t use this subcommand unless you have 2 economy providers set up."
```

NoahvdAa (commented):
Looks good, I'm not a big fan of some of the formatting though (for example: in the bukkit treasury command). Aside from some minor comments left below, LGTM

Geolykt:
The main issue I have right now is the naming of the "fork" package, which probably is just because It reminds me of [something else](https://linux.die.net/man/3/fork). Otherwise haven't looked into it tbh

bukkit/src/main/java/me/lokka30/treasury/plugin/bukkit/BukkitTreasuryPlugin.java
MrIvanPlays: yes. BungeeCord's ChatColor supports hex colors, while bukkit's doesn't.

api/src/main/java/me/lokka30/treasury/api/economy/EconomyProvider.java
```diff
@@ -17,34 +17,22 @@
...
 import java.util.UUID;
 
 /**
- * {@link Plugin Plugins} providing and managing economy data create a class
- * which implements this interface to be registered via the
- * {@link org.bukkit.plugin.ServicesManager ServicesManager} as a
- * {@link org.bukkit.plugin.RegisteredServiceProvider RegisteredServiceProvider&lt;EconomyProvider&gt;}.
+ * Implementors providing and managing economy data create a class
+ * which implements this interface to be registered in
+ * the specific platform they're implementing it for.
```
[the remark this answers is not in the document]
MrIvanPlays: Suggest alternative then.

bukkit/src/main/java/me/lokka30/treasury/plugin/bukkit/MessagesImpl.java
MrIvanPlays: The current implementation regenerates the file if it finds missing entries. That's why I added todo to find another way. I'm also thinking of [AnnotationConfig](https://github.com/MrIvanPlays/AnnotationConfig) which is able to handle missing options and config generation with comments all by itself. It needs more discussion.

bukkit/src/main/java/me/lokka30/treasury/plugin/bukkit/Treasury.java
```diff
@@ -88,6 +83,10 @@ public void onDisable() {
 
         // Add onDisable code here if required.
 
-        Utils.logger.info("&fShut-down complete (took &b" + shutdownTimer.getTimer() + "ms&f).");
+        logColor("&fShut-down complete (took &b" + shutdownTimer.getTimer() + "ms&f).");
+    }
+
+    private void logColor(String message) {
+        getLogger().info(ChatColor.translateAlternateColorCodes('&', message));
```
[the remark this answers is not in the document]
MrIvanPlays: oops, nice nit. thanks

bukkit/src/main/java/me/lokka30/treasury/plugin/bukkit/command/TreasuryCommand.java
MrIvanPlays: I tried to keep it as readable as possible.

core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/migrate/MigrateSubcommand.java
```diff
@@ -49,96 +51,106 @@
...
-                    new MultiMessage.Placeholder("prefix", main.messagesCfg.getConfig().getString("common.prefix"), true),
-                    new MultiMessage.Placeholder("providers", Utils.formatListMessage(main, new ArrayList<>(serviceProvidersNames)), false)
-            ));
+        if (args[0].equalsIgnoreCase(args[1])) {
+            sender.sendMessage(
+                    Message.of(
+                            MessageKey.MIGRATE_PROVIDERS_MATCH,
+                            placeholder("providers", Utils.formatListMessage(serviceProvidersNames))
+                    )
+            );
```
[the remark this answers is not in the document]
MrIvanPlays: It is making it long, hence problems with readability.

core/src/main/java/me/lokka30/treasury/plugin/core/config/messaging/MessageKey.java
```diff
@@ -0,0 +1,272 @@
...
+public enum MessageKey {
+    // Common
+    PREFIX(
+            "common.prefix",
+            new Messages.MessageHolder("&b&lTreasury:&7"),
+            ConfigVersion.V1
+    ),
+    NO_PERMISSION(
+            "common.no-permission",
+            new Messages.MessageHolder("%prefix% You don''t have access to that &8(&7requires permission &b%permission%&8)&7."),
```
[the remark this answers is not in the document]
MrIvanPlays: That's basically the defaults I copied from messages.yml

MrIvanPlays:
> The main issue I have right now is the naming of the "fork" package

suggest another name then. My first thought was "fork" because as you know "paper is a fork of spigot which is a fork of craftbukkit, the implementor of bukkit api, which is a fork of vanilla"

api/src/main/java/me/lokka30/treasury/api/economy/EconomyProvider.java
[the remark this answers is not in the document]
lokka30: I don't really see an issue with this

bukkit/src/main/java/me/lokka30/treasury/plugin/bukkit/command/TreasuryCommand.java
lokka30: I think this should be merged into one line but it is a nitpick, it's really fine

core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/migrate/MigrateSubcommand.java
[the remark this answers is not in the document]
lokka30: I think as a balance between the two we should have `sender.sendMessage` and `Message.of` on the same line. Barely less readable than the current one

bukkit/src/main/java/me/lokka30/treasury/plugin/bukkit/MessagesImpl.java
Jikoo: Personally would prefer something dependent on [SnakeYaml Engine](https://bitbucket.org/asomov/snakeyaml-engine/src/master/) (YAML 1.2 spec, built off SnakeYaml), but I haven't yet found a library that does that satisfactorily. Looks like Configurate plans to update eventually, though the issue for it has been pushed back twice. That may also solve the "yes" and "no" issues in paths, I believe a lot of SnakeYaml's original quirks were removed.

Jikoo:
> > The main issue I have right now is the naming of the "fork" package
> 
> suggest another name then. My first thought was "fork" because as you know "paper is a fork of spigot which is a fork of craftbukkit, the implementor of bukkit api, which is a fork of vanilla"

I agree that "fork" is not a great word, too much meaning. Maybe "vendor" as it's server vendor/implementation specific stuff?

Jikoo (approved):
Structurally sound, looks good to me.

I actually like the formatting, it seems much more readable as-is. Then again, I'm one of those people that tends to just run [google-java-format](https://github.com/google/google-java-format) over my projects every so often and call it a day.

I think problems that depend on library-related decisions are better left for other issues.

MrNemo64 (approved):
The only problem I see is the use of "fork" as others mentioned. Maybe changin it to "vendor" as suggested or "server_specific".
Just to point out, me.lokka30.treasury.api.economy.event.AccountTransactionEvent and me.lokka30.treasury.api.economy.event.AccountEvent seem to have a bit of an identation error

lokka30:
Thank you all for your input.

MrIvanPlays:
> Then again, I'm one of those people that tends to just run [google-java-format](https://github.com/google/google-java-format) over my projects every so often and call it a day.

Same I also use GJF but I haven't asked if lokka agrees on using that. Also this will require reformatting literally everything and I don't really think we _really_ need to change it.

Jikoo:
> > Then again, I'm one of those people that tends to just run [google-java-format](https://github.com/google/google-java-format) ([ 3 minutes ago](https://github.com/google/google-java-format/tree/35e97e8fae4d9474f94f7dd474fc3ba921dafa24)) over my projects every so often and call it a day.
> 
> Same I also use GJF but I haven't asked if lokka agrees on using that. Also this will require reformatting literally everything and I don't think we _really_ need to change it.

Oh yeah, no, not asking to do that here, more a comment on compactness vs readability. I vastly prefer the style you've got here to more compressed stuff.

lokka30 (approved):

MrIvanPlays:
let's discuss configuration ; what library shall we use? 

I'll tell more about [AnnotationConfig](https://github.com/MrIvanPlays/AnnotationConfig) ; it's a library which goal is to make configuration creation and access easier. Inspirations when creating it was [acf](https://github.com/aikar/commands) and velocity's 1.0.0 ways of generating their `velocity.toml`. 

It allows you to do custom serializers, even though you would probably never need it, because the default serializer does the trick in 90% of the cases. There's a special serializer and deserializer for enums and enum values. 

It allows you to have custom annotations with which you are able to validate a value or whatever you want of a field, annotated with it.

The built in annotations are: `Comment` for comments, `Ignore` to ignore something, `ConfigObject` is complex to explain with words so ill do it with code:

```java

@ConfigObject
private Something foo = new Something();

public static class Something {
  private String bar = "bar";
  private int baz = -1;
}

```

outputs this:
```yaml
foo:
  bar: "bar"
  baz: -1
```

of course it is possible with using a `Map<String, Object>` instead of such object, or an object such as bukkit's `Location`, for example ; 
`@Key` annotation to specify a different key ( by default the key is the field's name ) ; `@Min` and `@Max` to specify minimal or maximum value of a `String` and a number.

Example of generating treasury's config:
<details><summary>Settings.java</summary>

<p>
```java
@Comment("## Treasury")
@Comment("Treasury is a modern code library for SpigotMC plugins.")
// bla bla bla other header comments
public class Settings {

  @Comment("Settings regarding the update checker.")
  // bla bla bla comments
  @ConfigObject
  @Key("update-checker")
  private UCSettings updateCheckerSettings = new UCSettings();
  
  public static class UCSettings {
    // comments within sections support is yet to be added. 
    // idk, it may work with 2.0.1-SNAPSHOT of AnnotationConfig but
    // i don't guarantee it to work as it should so no @Comment here
    private boolean enabled = true;

    public boolean isEnabled() { 
      return enabled;
    }
  }
  
  @Comment("Settings regarding the debug logger.")
  // bla bla bla comments
  @ConfigObject
  @Key("debug")
  private DebugSettings debugSettings = new DebugSettings();
  
  public static class DebugSettings {
    
    @ConfigObject
    @Key("enabled-categories")
    private EnabledCategories enabledCategories = new EnabledCategories();
    
    public static class EnabledCategories {
      
      @Key("mode")
      // this is not the enum name we currently have its just an example
      private DebugMode debugMode = DebugMode.WHITELIST;
      
      @Key("list")
      private List<DebugCategory> categories = Collections.emptyList();
      
      public DebugMode getDebugMode() {
        return debugMode;
      }
      
      public List<DebugCategory> getCategories() {
        return categories;
      }
      
    }
    
    public EnabledCategories getEnabledCategories() {
      return enabledCategories;
    }
    
  }
  
  public UpdateCheckerSettings getUpdateCheckerSettings() {
    return updateCheckerSettings;
  }
  
  public DebugSettings getDebugSettings() {
    return debugSettings;
  }

}
```

</p>
</details>

The best is that we wouldn't need that `file` section it currently has, nor a config versioning system. 

Of course, AnnotationConfig has its downsides ; you can't do comments inside sections (yet) and the way you need to declare sections inside sections sucks. I'm thinking of making a resolver for `a.b.c` keys read from a `@Key` annotation which will transform the code demonstrated to be much cleaner.

That's what I am able to tell as of right now :) waiting for your feedback :)

lokka30:
Apart from the way sections are made, I like this. But I am still quite unsure about using it. I also want to see some alternatives just to make sure we adopt what's best.
Do the variables have to be private? I don't like how we have to create getters for each variable.

MrIvanPlays:
You can do sections in other ways, see the readme example. Fields can be public, but remember that java is an OOPL and however much you like it public fields are a sign of bad code. Getters are the proper way of interacting with something in an object, however much you don't like it. If you wanna use fields, go kotlin. They're called properties there and are much better to interact with than java's fields and getters.

MrIvanPlays:
Latest snapshot version of AnnotationConfig allows you to create sections such as:
```java
@Key("bar.baz.foo")
private String foo = "bar";
```
using comments even on this is still undefined behavior. I'm still searching for a feasible way to implement comment handling per value writing rather than it to be abstracted without breaking the API.

MrNemo64:
I've checked the annotation config and it looks interesting. I don't really know any config library but maybe we can try to have a general config in the core module, the core module doesn't know how to load/update the config, only knows what it should have so the loading and saving is the job of the child, this way we don't need a library we just use the default library system of the server. 
The child loads the library and gives it to the parent (maybe a Map<String, Object>), the parent saves it and checks that every key exists, if not adds its and gives it to the child to save it. However this has quite a lot of problems, for example loosing the comments in Bukkit/Spigot and it can be hard to implement so I don't know if it's worth. Just to mention the idea

MrIvanPlays:
show-off

Had to break 2 things to achieve this but coming soon :tm:

MrIvanPlays:
I'm still waiting on input. Tell me more about configurate. I've only heard about it, never really used it properly, only once and for test purposes. What other options do we have?

Geolykt:
I've only really heard of configurate. The big thing with configurate is it's HOCON support.

I'm a bit against annotation-based config gen because from my experience it's just magic, but as long as it isn't something that has to be forced on the implementation of the API I might as well not care about it.

MrIvanPlays:
> I'm a bit against annotation-based config gen because from my experience it's just magic

It's really not, reflection is not really magic. The only thing which is both WTF, OMG and MAGIC for me in java is the sun.misc.Unsafe class. 

> but as long as it isn't something that has to be forced on the implementation of the API

this PR is all about plugin implementation and nothing about the API. The only change we have to the API is that the base API is platform independent, hence removing the `getPlugin` method from `EcononomyProvider`, the only platform dependent API are events and such stuff.

Geolykt:
> It's really not, reflection is not really magic. The only thing which is both WTF, OMG and MAGIC for me in java is the sun.misc.Unsafe class.

Most petty human brains cannot properly debug reflection calls. Annotations make it even worse given that you are not actively (and knowingly) causing these reflection calls. A prime example is the `@EventHandler` annot. There are far too many instances where people forget it and wonder why they get the bugs they do. In this case the human debugger makes a false assumption. Of course I do not know whether this applies to your library as I haven't read any larger examples, nor do I have any experience with using it.

MrIvanPlays:
> Most petty human brains cannot properly debug reflection calls. Annotations make it even worse given that you are not actively (and knowingly) causing these reflection calls.

Reflection calls aren't _that_ bad as you describe them here. Reflection is quite useful in some circumstances. Problem comes when you abuse it, like with everything else. Yes, it is slower than you setting the fields manually, but who cares if your config loads 50-100ms slower? Or who cares if your command execution time is 50-100ms slower? The end user can never spot such a difference. And it is also not important at all to an application if it loads/executes something not as important slower when it can be faster, but with more or special code.

Annotations are a way for you to write handling logic once and apply it everywhere else. In our case, each plugin/java application has different settings, why have separate config handling logic for all of them when you can just have 1 and have it customizable?

> There are far too many instances where people forget it 

Newbies* . Newbies forget a lot of stuff and its not just about Bukkit. It's about java itself. I've never seen anyone who has at least basic knowledge of java and the Bukkit API miss that annotation. But let's not dig into that.

> this applies to your library

Can't really understand whether what applies to the library?

Geolykt:
> > Most petty human brains cannot properly debug reflection calls. Annotations make it even worse given that you are not actively (and knowingly) causing these reflection calls.
> 
> Reflection calls aren't _that_ bad as you describe them here. Reflection is quite useful in some circumstances. Problem comes when you abuse it, like with everything else. Yes, it is slower than you setting the fields manually, but who cares if your config loads 50-100ms slower? Or who cares if your command execution time is 50-100ms slower?

A human brain is a magnitude slower than a computer. Where a computer will take 10 μs a human brain might need minutes to reproduce **without any deviation** (with proper training you can get to far faster speeds, at the risk of errors). Unless we properly guard against unexpected events (such as illegal/unexpected configs, private constructors, inner classes, anonymous classes, JPMS, etc.) there is a risk with just assuming "it is annotated with x therefore y must be true". Most of the times such assertions are without any effect, other times they result in plugins breaking. Only rarely worse: security issues.

> The end user can never spot such a difference. And it is also not important at all to an application if it loads/executes something not as important slower when it can be faster, but with more or special code.

I am talking purely about developer perspective - and to some extent reliable and secure code - so it also affects users to a tiny extent. But yes, as long as everything is off-thread (which should be the case here) the user won't spot such a difference.

> Annotations are a way for you to write handling logic once and apply it everywhere else. In our case, each plugin/java application has different settings, why have separate config handling logic for all of them when you can just have 1 and have it customizable?

I cannot really disagree with this statement without repeating myself

> > There are far too many instances where people forget it
> 
> Newbies* . Newbies forget a lot of stuff and its not just about Bukkit. It's about java itself. I've never seen anyone who has at least basic knowledge of java and the Bukkit API miss that annotation. But let's not dig into that.

Well, I am one. Heck, even using my own bukkit-esque event handler I forget about the annotation

> > this applies to your library
> 
> Can't really understand whether what applies to the library?

Probably everything I listed so far. As of now it was more of a rant towards automatic dependency injection and annotations in general. Given that it appears that annotations are optional for your library my points may be void. On the other hand the name of it sends another sentiment.

---

While writing this comment I remember having heard of another config framework: https://github.com/heretere/hch It may be too bukkit-specific however

MrIvanPlays:
> Unless we properly guard against unexpected events

AnnotatedConfig does that. It has checks literally everywhere and if it finds such, it throws an exception.

> Well, I am one.

If you are developing using IntelliJ I suggest the Minecraft development plugin. It won't let you miss something because when you have a method annotated with `@EventHandler` it shows up that method's being used and if it's not it shows that it isn't. It also has other amazing features.

> Given that it appears that annotations are optional for your library my points may be void. On the other hand the name of it sends another sentiment.

Yes they are not requirement for a field to be dumped and loaded. I added "Annotation" in the name because it's based around annotations. Yes if you put a plain `private String foo = "bar";` it will get dumped and loaded. But it won't be as interesting as having comments, would it? So for that you have the `@Comment` annotation, and much more to alter the dumped configuration.

> While writing this comment I remembered having heard of another config framework

I understand about it just now. It is also using annotations. For me it is hard to use. E.g:
```java
ExampleAnnotatedConfig config = new ExampleAnnotatedConfig();
MultiConfigHandler configHandler = new MultiConfigHandler(Paths.get("."));
configHandler.load();
try {
  configHandler.loadConfigClass(config, ProcessorType.YAML);
} catch (/* cba to write exceptions */ e) {
 e.printStackTrace();
}
// ...
``` 

vs 
```java
File file = new File("config.yml");
ExampleAnnotatedConfig config = new ExampleAnnotatedConfig();
YamlConfig.getConfigResolver().loadOrDump(config, file);
// ...
```
This is the load/dump code that you will be doing 90% of the time with AnnotationConfig ^ . Of course if you are going to do custom annotations and custom serializers you would need some more code, but see how much less code that is? 

It is not bukkit oriented. It is modularized, with a special spigot part which registers bukkit-specific serializers. 

I appreciate your feedback about how you don't like annotations but I still want to convince you to take a look at not just AnnotationConfig but other annotation oriented libraries and not just for bukkit but for java as a whole. Take a look at sponge, spring boot, velocity, acf... Annotations are a very useful utility. I was like you ; I also hated them because "reflection", but I then got educated and I saw how useful and easy to use they were when I tried doing stuff with spring boot. 

Still need more feedback about how to generate/load the configuration and messages files. @lokka30 yours is the most important one so I'm still waiting for it.

MrIvanPlays:
If we're gonna talk speed ; I've just made a simple benchmark of AnnotationConfig if anyone's interested in the numbers. https://gist.github.com/MrIvanPlays/85d248a0e8ab15b71f45106e5359420d

lokka30:
Please could you show an example of accessing keys in a YAML file?

MrIvanPlays:
> Please could you show an example of accessing keys in a YAML file?

wdym? AnnotationConfig loads the config entries into the object you dump.

```java
private String foo = "bar";
```

will generate
```yaml
foo: "bar"
```

and if the user changes it to
```yaml
foo: "baz"
```

it's gonna be the same as:
```java
private String foo = "baz";
```

so you just access the field with a getter.

yannicklamprecht:
It's the way you do config normally. BukkitConfiguration is just the worst way of doing config.

lokka30:
> > Please could you show an example of accessing keys in a YAML file?
> 
> wdym? AnnotationConfig loads the config entries into the object you dump.
> 
> ```java
> private String foo = "bar";
> ```
> 
> will generate
> 
> ```yaml
> foo: "bar"
> ```
> 
> and if the user changes it to
> 
> ```yaml
> foo: "baz"
> ```
> 
> it's gonna be the same as:
> 
> ```java
> private String foo = "baz";
> ```
> 
> so you just access the field with a getter.

Oh, I apologize, I must have had a brainfart whilst asking that question.. Thanks. 😜

lokka30:
> It's the way you do config normally. BukkitConfiguration is just the worst way of doing config.

Do you have any thoughts of Ivan's use of AnnotationConfig in this PR?

lokka30:
Is there any way we can automate the getters for these - perhaps Lombok?

MrIvanPlays:
> Is there any way we can automate the getters for these - perhaps Lombok?

I don't really like lombok. I mean it is a cool library but it uses hacks. And also the fact a plugin for intellij is required so you can develop normally is just a no-go. 
Alt + Insert is much faster and better than lombok.

lokka30:
> > Is there any way we can automate the getters for these - perhaps Lombok?
> 
> I don't really like lombok. I mean it is a cool library but it uses hacks. And also the fact a plugin for intellij is required so you can develop normally is just a no-go. Alt + Insert is much faster and better than lombok.

I didn't know it required a plugin. A+I it is then 🙂

MrIvanPlays:
Not really _required_ but if you don't have it you're basically screwed writing that methods manually.

lokka30:
> Not really _required_ but if you don't have it you're basically screwed writing that methods manually.

Regardless it's a developer luxury that can be avoided. I don't think there will be many additions to the config files beyond what is currently there so I don't mind going the manual route - that is, IntelliJ's autocompletion.

Geolykt:
Lombok will result in the plugin being inaccessible for a lot of potential contributors - avoid it's use whenever possible

yannicklamprecht:
> > It's the way you do config normally. BukkitConfiguration is just the worst way of doing config.
> 
> Do you have any thoughts of Ivan's use of AnnotationConfig in this PR?


The only thing I remark ist that the PR should be split.
1. module split
2. Config stuff
3. probably other stuff that are own pr's

3k LOC aren't well reviewable for me.

MrIvanPlays:
Nono @yannicklamprecht he asks whether or not you approve to replace the homebrew config ( which currently sucks ) with AnnotationConfig.

yannicklamprecht:
> Nono @yannicklamprecht he asks whether or not you approve to replace the homebrew config ( which currently sucks ) with AnnotationConfig.

From the frontend standpoint I'm fine with that new config design.
That stuff that is currently implemented in this pr has a lot of overhead with the whole key thingy.
When stuff can be easily maintained I'll be fine.

MrIvanPlays:
I think that the codemc deployment profile should not be done by me so it is gonna be left as a TODO.

Waiting for your 2 cents on the code :)

MrNemo64 (approved):

lokka30:
Suggestion 1:
I think we should rename the `bukkit` module to `core-bukkit`

Suggestion 2:
At: `bukkit\src\main\java\me\lokka30\treasury\plugin\bukkit\vendor\paper\PaperAsyncTabEnhancement.java` Line 47, and
At: `core\src\main\java\me\lokka30\treasury\plugin\core\command\subcommand\migrate\MigrateSubcommand.java` Line 188:
We should instead iterate over the plugins which have registered as a Treasury service provider. Is that still possible?

Suggestion 3:
At: `bukkit\src\main\java\me\lokka30\treasury\plugin\bukkit\vendor\paper\PaperEnhancements.java` Line 10:
Wouldn't it be more reliable to check versions the way MicroLib does? (e.g. to check if the server is 1.12+, ensure `Material.WHITE_CONCRETE` exists. to check if the server is 1.15+, check if `Material.HONEY_BLOCK` exists)

Suggestion 4:
At: `core\src\main\java\me\lokka30\treasury\plugin\core\command\subcommand\InfoSubcommand.java` Line 54, 55:
We can probably replace this with `"${project.version}` and `${project.description}`

Suggestion 5:
At: `bukkit\src\main\java\me\lokka30\treasury\plugin\bukkit\Treasury.java`
I think this class should be renamed to `TreasuryBukkit`

Suggestion 6:
At: `core\src\main\java\me\lokka30\treasury\plugin\core\config\messaging\Messages.java` Line 31:
Forgot a `m` in `.yl` (to make `.yml`)

MrIvanPlays:
> Suggestion 1: I think we should rename the `bukkit` module to `core-bukkit`

`core` implies that it doesn't have the full functionality of the plugin. In my opinion that's a no.

> Suggestion 3: At: `bukkit\src\main\java\me\lokka30\treasury\plugin\bukkit\vendor\paper\PaperEnhancements.java` Line 10: Wouldn't it be more reliable to check versions the way MicroLib does? (e.g. to check if the server is 1.12+, ensure `Material.WHITE_CONCRETE` exists. to check if the server is 1.15+, check if `Material.HONEY_BLOCK` exists)

The best method is what it's currently using as it is guaranteed the craftbukkit package is gonna contain the version. No, materials are the worst way to check the versions because they're subject of a rapid change between releases.

> Suggestion 4: At: `core\src\main\java\me\lokka30\treasury\plugin\core\command\subcommand\InfoSubcommand.java` Line 54, 55: We can probably replace this with `"${project.version}` and `${project.description}`

Maven does not handle class filtering unless you have a maven plugin in the build and put the classes you need filtered in a special directory (a.ka `src/main/java-templates` or something, don't really remember). So no, we can't replace them. Either way you have the version replaced in the plugin.yml, and the current bukkit platform implementation code is getting them from there. Other platforms also have something like plugin.yml which is holding the version, so we're safe with the current implementation for now.

lokka30:
Is this PR ready to be merged?

lokka30:
@Jikoo, do you have any comments? I understand you are likely busy so don't worry about it if you can't.

Jikoo (approved):
I usually prefer using `dependencyManagement` and `pluginManagement` to properties for versioning in multi-module configurations, but it works fine and I'm okay with the current state. The restructure seems good, formatting is solid.

I apologize that I haven't looked at any of the configuration details, I spent most of my programming time recently prepping for 1.18.

MrIvanPlays:
> Is this PR ready to be merged?

If no one has anything else to say then yes.

lokka30:
Merged - thank you very much Ivan and also to everyone who provided feedback on the PR!

PR #51, opened 2021-12-06
Rework currencies

### LINKED ISSUE #40: Parsing sums of money
Unlike vault, Treasury provides multi-currency support. Given that this multi-currency support is a thing in the backend, API users may be tempted to expose currencies to the frontend. Implementing `/balance` is implemented easily, albeit expensively by iterating over each and every currency and doing other trivial operations. However commands such as `/pay` would be a lot more complicated to implement as now you'd have to parse the amount.
It is rather easy if you have a syntax that operates like `/pay <reciever> <amount> [currency]`, where as currency has to be a currency name; easy enough to implement given that the currencies name can be looked up easily via Treasury API. However what if the User wants to format the amount to something like `15€`? As far as I see there is no sane API to resolve that currency. The same applies to `$ 15`, `$15`, `15 USD` or `15 €`. An API user would not like to do this parsing hell manually so I propose an API to allow parsing strings such as the previous examples. This is aided by the fact that Currencies are only provided by the economy provider, so 3rd party currencies would not prove to be much of an issue.

What would the best course of action be here?

### LINKED ISSUE #49: Entirely rework currencies
I've started doing javadocs and stuff, and I got to start documenting the currency conversion system.

The currency conversion design flaws I've seen so far

1. Using hash codes
2. Requiring a registration, even though it still needs a currency conversion to be passed if you want to convert a specific value

In my opinion, this API shouldn't exist at all. I mean, if you want to jump through currencies, there should be another way to do that.

As for the currency interface: it shouldn't be an interface after all, but rather an immutable object which holds the data this interface holds, and some sort of a CurrencyFormatter interface and some sort of a CurrencyStartingBalanceGetter or something.

MrIvanPlays:
Closes #49 
Closes #40


MrIvanPlays:
@MrNemo64 @Geolykt

api/src/main/java/me/lokka30/treasury/api/economy/currency/CurrencyParseResult.java
```diff
@@ -0,0 +1,123 @@
...
+    /**
+     * Creates a new {@link CurrencyParseResult}, which is not indicating successful parse.
+     *
+     * @param state the state/reason why parsing failed
+     * @return state only parse result
+     * @author MrIvanPlays
+     * @since v1.0.0
+     */
+    @NotNull
+    public static CurrencyParseResult stateOnly(@NotNull State state) {
```
lokka30, line 74: Perhaps this should be renamed to `failure`

api/src/main/java/me/lokka30/treasury/api/economy/currency/CurrencyParseResult.java
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    private final State state;
+    private final OptionalDouble value;
+    private final Optional<Currency> currency;
+
+    private CurrencyParseResult(@NotNull State state, double value, Optional<Currency> currencyOptional) {
+        this.state = Objects.requireNonNull(state, "state");
+        this.value = value == -1 ? OptionalDouble.empty() : OptionalDouble.of(value);
+        // I put this comment here specially for lokka30: ignore this warning this is exactly what it shall be
```
lokka30, line 85: not necessary

api/src/test/java/me/lokka30/treasury/api/economy/currency/CurrencyTests.java
```diff
@@ -0,0 +1,66 @@
...
+        Assertions.assertEquals('€', parseCurrency("€1"));
+        Assertions.assertEquals('€', parseCurrency("1€"));
+        Assertions.assertEquals('€', parseCurrency("euro1"));
+    }
+
+    private Character parseCurrency(String input) {
+        Currency currency = CurrencyManager.INSTANCE.parseCurrencyAndValue(input).getCurrency().orElse(null);
+        return currency == null ? null : currency.getCurrencyCharacter();
+    }
+}
```
lokka30, line 66: This shouldn't be part of the API, rather, part of the core module, used for debugging purposes.

lokka30 (changes requested):

lokka30:
Thank you very much! Only a few things I could point out to change, the rest is marvelous.

api/src/test/java/me/lokka30/treasury/api/economy/currency/CurrencyTests.java
MrIvanPlays: This is a test. Tests are not included in deployment, neither in the api, neither in the distributed plugin jar.

lokka30:
Would like to hear from @Jikoo @MrNemo64 @Geolykt, if possible. Would be much appreciated :)

api/src/test/java/me/lokka30/treasury/api/economy/currency/CurrencyTests.java
lokka30: Ah, thanks for pointing that out, I never noticed it was in the test folder.

lokka30 (approved):

lokka30:
Once we've gathered enough feedback, is this PR ready to merge as-is?

MrIvanPlays:
> Once we've gathered enough feedback, is this PR ready to merge as-is?

Yes.

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
```diff
@@ -12,69 +12,224 @@
...
  * currencies for different transactions, e.g. a daily reward
  * plugin can award players 'Tokens', but a job plugin
  * can award players 'Dollars'. Facilitates great customisability.
+ *
+ * @author lokka30, Geolykt, MrIvanPlays
+ * @since v1.0.0
  */
-@SuppressWarnings("unused")
-public interface Currency {
+public final class Currency {
```
MrNemo64, line 33: I don't think this class should be final as I explained here. 
https://github.com/lokka30/Treasury/issues/49#issuecomment-986275101
With the example of simulating money, lets say my plugin simulates cryptocurrencyes:
I have bitcoin and etherium, depening on how much is buyed/selled i want these currencyes to have a higer/lower value, i can modify these by changing its conversionCoefficient, I could do this in several ways:
1- Override currency so I can change the returned value as I want
2- Change the value of conversionCoefficient when necesary (its final so it's kinda hard)
3- conversionCoefficient is a supplier that I provide (?)
Right now I couldn't simulate this because I can't modify the conversionCoefficient

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
```diff
@@ -12,69 +12,224 @@
...
+                     double conversionCoefficient,
+                     @Nullable Function<UUID, Double> startingBalance,
+                     @NotNull BiFunction<Double, Locale, String> balanceFormatter,
+                     @NotNull String... names
+    ) {
+        if (currencyId == null) {
+            this.currencyId = UUID.randomUUID();
+        } else {
+            this.currencyId = currencyId;
+        }
```
MrNemo64, line 119: Just currios, why not
`this.currencyId = currencyId == null ? UUID.randomUUID() : currencyId;`

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
```diff
@@ -12,69 +12,224 @@
...
+
+    /**
+     * Gets the character which identifies this currency, which may be null.
+     *
+     * @return char identifier. could be null
+     * @author MrIvanPlays
+     * @since v1.0.0
+     */
+    @Nullable
+    public Character getCurrencyCharacter() {
```
MrNemo64, line 155: Maybe this should be a String? Maybe the currency symbol has several characters

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
```diff
@@ -12,69 +12,224 @@
...
+        if (startingBalance == null) {
+            this.startingBalance = ($) -> 0d;
+        } else {
+            this.startingBalance = startingBalance;
+        }
+        if (names.length < 1) {
+            throw new IllegalArgumentException("Empty array specified for currency name.");
+        }
+        this.balanceFormatter = Objects.requireNonNull(balanceFormatter, "balanceFormatter");
+        this.names = Objects.requireNonNull(names, "names");
```
MrNemo64, line 132: The throws and Objects.requireNonNull could be at the begining of the constructor

api/src/main/java/me/lokka30/treasury/api/economy/currency/CurrencyManager.java
```diff
@@ -0,0 +1,248 @@
...
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Represents a manager of {@link Currency currencies}.
+ *
+ * @author MrIvanPlays
+ * @since v1.0.0
+ */
+public enum CurrencyManager {
+    INSTANCE;
```
MrNemo64, line 17: Shoudln't this be `CurrencyManager INSTANCE;` ?

api/src/main/java/me/lokka30/treasury/api/economy/currency/CurrencyManager.java
```diff
@@ -0,0 +1,248 @@
...
+     *
+     * @param from  the currency to convert from
+     * @param to    the currency to convert to
+     * @param value the value to convert
+     * @return converted value
+     * @throws IllegalArgumentException if one or both of the currencies are not registered
+     * @author MrIvanPlays, MrNemo64
+     * @since v1.0.0
+     */
+    public double convertCurrency(@NotNull Currency from, @NotNull Currency to, double value) {
```
MrNemo64, line 173: Could also have a 
```java
public double convertCurrency(@NotNull UUID fromId, @NotNull UUID toId, double value) {
  Optional<Currency> from = getCurrency(fromId);
  Optional<Currency> to = getCurrency(toId);
  if(from.isEmpty() || to.isEmpty())
    throw new IllegalArgumentException("some mesage");
  return convertCurrency(from.get(), to.get(), value);
}
```

And maybe instead of value it should be called amount? Dosnt really matter

MrNemo64 (changes requested):

api/src/main/java/me/lokka30/treasury/api/economy/currency/CurrencyManager.java
```diff
@@ -0,0 +1,248 @@
...
+        Objects.requireNonNull(from, "from");
+        Objects.requireNonNull(to, "to");
+        if (!hasCurrency(from.getCurrencyId()) || !hasCurrency(to.getCurrencyId())) {
+            throw new IllegalArgumentException("Specified currencies are unknown to the currency manager");
+        }
+        if (from.getConversionCoefficient() == 0 || to.getConversionCoefficient() == 0 || value == 0) {
+            return 0;
+        }
+        // thanks MrNemo64 for the formula
```
MrNemo64, line 182: Pun on a comment or something https://github.com/lokka30/Treasury/issues/49#issuecomment-986255592 so if anyone more doesn't understand the formula he can just read the explanation

MrNemo64 (changes requested):

api/src/main/java/me/lokka30/treasury/api/economy/currency/CurrencyManager.java
MrIvanPlays: I decided to use enum cuz that's the only way you can _really_ enforce 1 instance of a file. To answer your question, no.

api/src/main/java/me/lokka30/treasury/api/economy/currency/CurrencyManager.java
MrIvanPlays: > And maybe instead of value it should be called amount? Dosnt really matter

as you said, doesn't really matter.

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
MrIvanPlays: it is not necessary to specify it anyway so it is staying as `char` due to that the most used global currencies are using characters (e.g. '$' is the dollar, '€' is the euro)

api/src/main/java/me/lokka30/treasury/api/economy/currency/CurrencyManager.java
MrIvanPlays: done.

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
MrIvanPlays: done.

api/src/main/java/me/lokka30/treasury/api/economy/currency/CurrencyManager.java
MrIvanPlays: added.

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
```diff
@@ -12,69 +12,216 @@
...
+     */
+    @NotNull
+    public static Currency of(
+            @NotNull UUID currencyId,
+            @Nullable Character currencyChar,
+            int roundedDigits,
+            double conversionCoefficient,
+            @Nullable Function<UUID, Double> startingBalance,
+            @NotNull BiFunction<Double, Locale, String> balanceFormatter,
+            @NotNull String... names
```
Jikoo, line 85: I forget if it works on varargs, but you may consider making these `@NotNull String @NotNull ... names` instead to signal that null elements are also disallowed.

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
```diff
@@ -12,69 +12,216 @@
...
+    ) {
+        Objects.requireNonNull(currencyId, "currencyId");
```
Jikoo, line 87: Aren't all of these strictly redundant? I'm almost certain the JetBrains `@NotNull` annotation is processed during compile to include the required validations.

api/src/main/java/me/lokka30/treasury/api/economy/currency/CurrencyManager.java
MrNemo64: Seems kinda hacky. 

```java
public static final CurrencyManager INSTANCE = new CurrencyManager();

private CurrencyManager() {
  if(INSTANCE != null) throw new IllegalCallerException("Tryed to create more than one instances of CurrencyManager");
}

```
Doesn't this work?

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
MrNemo64: I suppose that if someone wants to use a mylti character symbol he can just use a name

core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/InfoSubcommand.java
```diff
@@ -70,7 +70,7 @@ public void execute(@NotNull CommandSource sender, @NotNull String label, @NotNu
                     placeholder("api-version", provider.getSupportedAPIVersion()),
                     placeholder("supports-bank-accounts", Utils.getYesNoStateMessage(provider.hasBankAccountSupport())),
                     placeholder("supports-transaction-events", Utils.getYesNoStateMessage(provider.hasTransactionEventSupport())),
-                    placeholder("primary-currency", provider.getPrimaryCurrency().getCurrencyName()))
+                    placeholder("primary-currency", provider.getPrimaryCurrency().getPrimaryCurrencyName()))
```
MrNemo64, line 73: maybe add as placeholder "currency_<uuid>_name_<num>" to use a spacific name of a curency?

MrIvanPlays: primary name should be shown as the other names may be only for parse ease
e.g. `dollar, dolla, dola, buck, bucks`

api/src/main/java/me/lokka30/treasury/api/economy/currency/CurrencyManager.java
MrIvanPlays: you can't have anything in the constructor. this code at the point of the constructor call `INSTANCE` is already non null, leading to the exception. 

enforcing a single instance is something full of compromises, and as I said the only way to _really_ enforce a single instance is by using an enum. if we use another method, someone with bad intentions can break all the currencies and leave permanent damage on the economy.

core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/InfoSubcommand.java
MrNemo64: oh, I see

Jikoo (approved):
Barring the potential annotation-related changes, looks good to me 👍

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
MrIvanPlays: no. jetbrains' annotations are only a javadoc perk and a perk for intellij users so it bugs them to check nullability.

api/src/main/java/me/lokka30/treasury/api/economy/currency/CurrencyManager.java
MrNemo64:
```java
public class HelloWorld{

     public static void main(String []args){
        Singleton.INSTANCE.sayHi();
     }
     
     static class Singleton {
        public static final Singleton INSTANCE = new Singleton();
  
          private Singleton() {
            if(INSTANCE != null)
        	  throw new RuntimeException("Tryed to create more than one instances of Singleton");
          }
          
          public void sayHi() {
        	System.out.println("Hi!");
          }
  
    }
     
}
```
This runs without problems, I think I'm not understanding you

Jikoo: Using a single element enum isn't super common practice but there's nothing inherently wrong with it compared to more traditional singleton patterns. Theoretically it's roughly the equivalent of any other similar security measures - it effectively is a final class with a public final field pre-instantiated. A sufficiently determined person could still replace it, but they'd have to really be trying. It just saves on writing boilerplate code to create the instance.

MrIvanPlays: > A sufficiently determined person could still replace it, but they'd have to really be trying

to add to this, consider we have this code: 
```java
public class DoesntMatter {

  private static DoesntMatter instance;

  public static DoesntMatter getInstance() {
    if (instance == null) {
      instance = new DoesntMatter();
    }
    return instance;
  }

  private DoesntMatter() {
  }

  // whatever

}
```

the `instance` field can easily be altered via reflection:
```java
Field field = DoesntMatter.class.getDeclaredField("instance");
field.setAccessible(true);
Constructor constructor = DoesntMatter.class.getDeclaredConstructor();
field.set(null, constructor.newInstance());
```

there I just altered the instance of that class ^, potentially clearing all the maps and all the stuff this class needs to function properly during runtime.

whilst as Jikoo said with an enum they should really be trying, but with such designs - 4 lines and I'm able to ruin the entire plugin (make it spit errors, do whatever).

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
Jikoo: https://www.jetbrains.com/help/idea/nullable-and-notnull-annotations.html

> When you compile your project, the IDE adds assertions to all methods and parameters annotated with the `@NotNull` annotation. The assertions will fail if null is passed in code where `@NotNull` is expected. You can disable this option and configure the list of annotations in the Settings/Preferences dialog Ctrl+Alt+S. Go to Build, Execution, Deployment | Compiler.

Looks like it's not just syntactical sugar to me, but I'm not sure if that's purely on IntelliJ's side. I know for a fact I've had the checks generate while compiling with Maven, however, that was inside IntelliJ.

api/src/main/java/me/lokka30/treasury/api/economy/currency/CurrencyManager.java
MrNemo64: I see

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
MrIvanPlays: Just compiled via the terminal:
decompiler-source

do you see anything generated? I may be blind or idk, but I don't see it.
Irony aside, the plugin is going to be build and deployed to maven repositories via a CI and not by intellij/human, so if you want that code to generate not just if you compile via intellij, you need to find a maven plugin which does it and put it. But that has nothing to do with this PR.

MrNemo64 (approved):

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
Jikoo: Yeah, looks like it's IJ-specific. I'll have to look into it more, would be useful to have.

lokka30:
I'll leave about a day until I merge it in case @Geolykt or anyone else wants to review the PR, also leaves more time in case you find an issue that needs to be addressed.

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
```diff
@@ -12,69 +12,216 @@
...
+                balanceFormatter,
+                names
+        );
+    }
+
+    private final UUID currencyId;
+    private final Character currencyChar;
+    private final String[] names;
+    private final int roundedDigits;
+    private final double conversionCoefficient;
```
MrNemo64, line 103: Can we make this protected and not final? Or a protected setter?

MrIvanPlays: I don't understand why. I don't even understand why you want to extend Currency. Sure thing, you want to do cryptocurrencies and you for some reason need to change the conversion coefficient. Just do it via a config of some kind and make your server restart every 12 hours to apply the changes or every day at 12 midnight. 

The given value is and shall not be controlled by the conversion coefficient, unless you _really_ need to convert from currency x to currency y. You can fetch every day from the internet the value of the bitcoin and do calculations based on that in order to determine what amount shall you give them. 

I'm starting to think to return the `final` modifier on the class too...

Jikoo: Why make end users restart to reload a configuration? That will only encourage use of `/reload`. There's also no real harm in allowing extension of currencies, though I agree it doesn't seem like a particularly great thing to do. Allowing on-the-fly currency modification would definitely make it more possible to do stock exchange style things without having to unregister (which, as much as there's no API for it, is certainly possible with reflection) and re-register the currency.

MrIvanPlays: @Jikoo @MrNemo64 happy now!?

api/src/main/java/me/lokka30/treasury/api/economy/currency/CurrencyManager.java
```diff
@@ -65,6 +65,102 @@ public void registerCurrency(@NotNull Currency currency) {
...
+        for (Currency currency : currencies) {
+            registerCurrency(currency);
+        }
+    }
+
+    /**
+     * @param currencies currencies to register
+     * @see #registerCurrency(Currency)
+     */
+    public void registerCurrency(@NotNull Iterable<@NotNull Currency> currencies) {
```
lokka30, line 83: Should this be `registerCurrencies`?

api/src/main/java/me/lokka30/treasury/api/economy/currency/CurrencyManager.java
```diff
@@ -65,6 +65,102 @@ public void registerCurrency(@NotNull Currency currency) {
...
+        if (currency.getCurrencyCharacter() != null) {
+            BY_CHAR.remove(currency.getCurrencyCharacter());
+        }
+    }
+
+    /**
+     * @param currencies the currencies you want unregistered
+     * @see #unregisterCurrency(Currency)
+     */
+    public void unregisterCurrency(@NotNull Currency @NotNull ... currencies) {
```
lokka30, line 112: Should this be `unregisterCurrencies`?

api/src/main/java/me/lokka30/treasury/api/economy/currency/CurrencyManager.java
```diff
@@ -65,6 +65,102 @@ public void registerCurrency(@NotNull Currency currency) {
...
+     */
+    public void unregisterCurrency(@NotNull UUID currencyId) {
+        getCurrency(currencyId).ifPresent(this::unregisterCurrency);
+    }
+
+    /**
+     * @param currencyIds currency ids to unregister
+     * @see #unregisterCurrency(UUID)
+     */
+    public void unregisterCurrency(@NotNull UUID @NotNull ... currencyIds) {
```
lokka30, line 146: Should this be `unregisterCurrencies`?

api/src/main/java/me/lokka30/treasury/api/economy/currency/CurrencyManager.java
```diff
@@ -65,6 +65,102 @@ public void registerCurrency(@NotNull Currency currency) {
...
+        for (UUID currencyId : currencyIds) {
+            unregisterCurrency(currencyId);
+        }
+    }
+
+    /**
+     * @param currencyIds currency ids to unregister
+     * @see #unregisterCurrency(UUID)
+     */
+    public void unregisterCurrency(@NotNull Iterable<@NotNull UUID> currencyIds) {
```
lokka30, line 157: Should this be `unregisterCurrencies`?

lokka30 (changes requested):

api/src/main/java/me/lokka30/treasury/api/economy/currency/CurrencyManager.java
MrIvanPlays: why should it be

api/src/main/java/me/lokka30/treasury/api/economy/currency/CurrencyManager.java
MrIvanPlays: why should it be

api/src/main/java/me/lokka30/treasury/api/economy/currency/CurrencyManager.java
MrIvanPlays: why should it be

api/src/main/java/me/lokka30/treasury/api/economy/currency/CurrencyManager.java
MrNemo64: Because it allows to register several currencies at ones

api/src/main/java/me/lokka30/treasury/api/economy/currency/CurrencyManager.java
MrNemo64: Because it allows to unregister several currencies at ones.
I do prefer the name we have now

api/src/main/java/me/lokka30/treasury/api/economy/currency/CurrencyManager.java
MrNemo64: Because it allows to unregister several currencies at ones

MrIvanPlays:
@lokka30 Happy now!?

MrNemo64:
If we rename one of the methods that remove currencies by id, shouldn't all the methods that remove by id also change to match the style of the name? I know is not really necessary but to keep consistency

MrIvanPlays:
> If we rename one of the methods that remove currencies by id, shouldn't all the methods that remove by id also change to match the style of the name? I know is not really necessary but to keep consistency

And I think to myself........ Why the fuck

lokka30 (approved):

lokka30:
> @lokka30 Happy now!?

They were just very minor nitpicks, I was happy to begin with :)
Will merge now.



> > If we rename one of the methods that remove currencies by id, shouldn't all the methods that remove by id also change to match the style of the name? I know is not really necessary but to keep consistency
> 
> And I think to myself........ Why the fuck

Consistency can be quite important, we should apply it wherever we can. If it's unfeasible for us to modify something to strive for better consistency then that's completely fine. MrNemo said it wasn't important so we should focus on the next things to do for Treasury instead of trying to iron out every single detail for the moment

PR #52, opened 2021-12-07
Versioning changes

### LINKED ISSUE #46: Changes to the EconomyAPIVersion
Right now the EconomyAPIVersion class is really simple: Identifyes mayor versions with a number, not more not less. I think this should change. I propose some changes

1. Change the use of the `@since` tag, instead of `@since v1.0.0` we could use `@since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#VERSION_1 v1.0.0}` in the documentation shows the same but it provides a link to the enum version
2. Change enum entryes naming, instead of `VERSION_1` we could use `v1_0_0` 
3. Specify an order. All enums implement the `Comparable<>` interface with the default logic being indexes. Since this logic doesn't necesarily apply to us it would make sense to specify out comparing logic
4. Change the number system. Using integers to know what version it is means that we can only represent big updates, we could have an array of ints/shorts/bytes (probably shorts is enough for all versions we may release)
5. (Optional) in de documentation of each entry list the changes of that version

I can make a class with this features so you see what I mean

### LINKED ISSUE #48: Warn the server console if an implementation is using a older/newer Treasury API version
We can use `ServiceRegisterEvent` to check when one has registered, then just run some logic.

MrIvanPlays:
Partly closes #46 
Closes #48

### DIFF
--- a/api/src/main/java/me/lokka30/treasury/api/economy/EconomyProvider.java
+++ b/api/src/main/java/me/lokka30/treasury/api/economy/EconomyProvider.java
@@ -36,9 +36,12 @@ public interface EconomyProvider {
     /**
      * Get the version of the Treasury API the {@code EconomyProvider} is based on.
      *
-     * <p>Please note that the Treasury API version is not the same as the Spigot API version!
+     * <p>Warning: The Treasury API version is completely different to any other platform version,
+     * such as the 'api-version' value in the Bukkit implementation's plugin.yml file.
+     * <p>Warning: Do not use {@link EconomyAPIVersion#getCurrentAPIVersion()}, this is for internal Treasury use only.
+     * <b>You must only use the constants provided.</b>
      *
-     * @author lokka30
+     * @author lokka30, MrIvanPlays
      * @return the API version
      * @since v1.0.0
      */
--- a/api/src/main/java/me/lokka30/treasury/api/economy/misc/EconomyAPIVersion.java
+++ b/api/src/main/java/me/lokka30/treasury/api/economy/misc/EconomyAPIVersion.java
@@ -12,14 +12,57 @@
 
 package me.lokka30.treasury.api.economy.misc;
 
-@SuppressWarnings("unused")
+/**
+ * Represents the current API version.
+ *
+ * @author lokka30, MrIvanPlays
+ * @since v1.0
+ */
 public enum EconomyAPIVersion {
+    v1_0(new short[] {1, 0})
+    ;
 
-    VERSION_1(1);
+    /**
+     * Returns the current economy api version.
+     *
+     * @return current version
+     * @deprecated internal use only
+     */
+    @Deprecated
+    public static EconomyAPIVersion getCurrentAPIVersion() {
+        EconomyAPIVersion[] constants = EconomyAPIVersion.values();
+        return constants[constants.length - 1];
+    }
 
-    private final int number;
-    EconomyAPIVersion(final int number) { this.number = number; }
+    private final short[] versionArray;
 
-    public int getNumber() { return number; }
+    EconomyAPIVersion(short[] versionArray) {
+        if (versionArray == null || versionArray.length > 2) {
+            throw new IllegalArgumentException("versionArray must not be null and must contain only 2 versions e.g 1, 0");
+        }
+        this.versionArray = versionArray;
+    }
+
+    /**
+     * Returns the major api version as a short value for this {@link EconomyAPIVersion}
+     *
+     * @return major api version
+     * @author MrIvanPlays
+     * @since v1.0
+     */
+    public short getMajorRevision() {
+        return versionArray[0];
+    }
+
+    /**
+     * Returns the minor api version as a short value for this {@link EconomyAPIVersion}
+     *
+     * @return minor api version
+     * @author MrIvanPlays
+     * @since v1.0
+     */
+    public short getMinorRevision() {
+        return versionArray[1];
+    }
 
 }
--- a/core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/InfoSubcommand.java
+++ b/core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/InfoSubcommand.java
@@ -13,6 +13,7 @@
 package me.lokka30.treasury.plugin.core.command.subcommand;
 
 import me.lokka30.treasury.api.economy.EconomyProvider;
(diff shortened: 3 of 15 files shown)


lokka30:
Thanks Ivan! Is this PR still a draft, or is it ready to be merged (once reviewed)?

api/src/main/java/me/lokka30/treasury/api/economy/misc/EconomyAPIVersion.java
```diff
@@ -12,14 +12,55 @@
...
 public enum EconomyAPIVersion {
+    v1_0(new short[] {1, 0})
+    ;
+
+    /**
+     * Returns the latest economy api version.
+     *
+     * @return latest economy api version
+     */
+    public static EconomyAPIVersion getLatestVersion() {
```
lokka30, line 30: I believe `getCurrentVersion()` is a better name, because the user could have a very old version of Treasury installed so an economy provider could think that Treasury is providing the latest one available online.

core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/InfoSubcommand.java
```diff
@@ -54,7 +55,7 @@ public void execute(@NotNull CommandSource sender, @NotNull String label, @NotNu
                 placeholder("version", main.getVersion()),
                 placeholder("description", main.getDescription()),
                 placeholder("credits", "https://github.com/lokka30/Treasury/wiki/Credits"),
-                placeholder("latest-api-version", main.getEconomyAPIVersion()),
+                placeholder("latest-api-version", EconomyAPIVersion.getLatestVersion()),
```
lokka30, line 58: Same comment here as above, I think this should be renamed to `current-api-version` for the same reason

core/src/main/java/me/lokka30/treasury/plugin/core/utils/PluginVersion.java
```diff
@@ -0,0 +1,151 @@
...
+        if (input.indexOf('.') == -1) {
+            throw new IllegalArgumentException(
+                    "Illegal Treasury version found! " +
+                            "If this is a live environment version or you're a developer doing changes, please fix this ASAP!"
+            );
+        }
+        String[] parts = input.split("\\.");
+        if (parts.length != 3) {
+            throw new IllegalArgumentException(
+                    "Illegal Treasury version found! " +
```
lokka30, line 37: Since this String is used twice, should we store it somewhere?

core/src/main/java/me/lokka30/treasury/plugin/core/utils/UpdateChecker.java
```diff
@@ -38,18 +41,102 @@ public static void checkForUpdates() {
...
-                    if (!plugin.getVersion().equalsIgnoreCase(latestVersion)) {
-                        plugin.logger().warn(
-                                "A new Treasury update is available - '&bv"
-                                        + latestVersion
-                                        + "&7' - please update as soon as possible."
-                                        + " &8(&7You're running '&bv" + plugin.getVersion() + "&7'&8)"
-                        );
-                    }
+                    handleVersionCheck(latestVersion);
                 }
```
lokka30, line 45: We should also catch `NetException` (if I remember  correctly), which fires if the user's server couldn't connect

lokka30 (changes requested):
Thank you - very good! Here are the only things I could pick out from the PR.

lokka30:
Once the changes have been resolved, is the PR still in a draft state, or ready to be merged? Would you like others to review this too?

api/src/main/java/me/lokka30/treasury/api/economy/misc/EconomyAPIVersion.java
```diff
@@ -12,14 +12,55 @@
 
 package me.lokka30.treasury.api.economy.misc;
 
-@SuppressWarnings("unused")
+/**
+ * Represents the current API version.
+ *
+ * @author lokka30, MrIvanPlays
+ * @since v1.0
```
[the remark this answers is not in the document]
MrIvanPlays: economy api version shall be like that instead of the link.

api/src/main/java/me/lokka30/treasury/api/economy/misc/EconomyAPIVersion.java
```diff
@@ -12,14 +12,55 @@
...
+ */
 public enum EconomyAPIVersion {
```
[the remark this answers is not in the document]
MrIvanPlays: unnecessary bloat

core/src/main/java/me/lokka30/treasury/plugin/core/utils/PluginVersion.java
MrIvanPlays: no need.

core/src/main/java/me/lokka30/treasury/plugin/core/utils/UpdateChecker.java
MrIvanPlays: nowhere in java's documentation specifies that anything like that gets thrown. fairly sure catching only the IOException is safe.

api/src/main/java/me/lokka30/treasury/api/economy/EconomyProvider.java
```diff
@@ -36,7 +36,11 @@
     /**
```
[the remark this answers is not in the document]
MrIvanPlays: done.

MrIvanPlays:
> Would you like others to review this too?

there's no need for others to review as the code is pretty straightforward.

api/src/main/java/me/lokka30/treasury/api/economy/misc/EconomyAPIVersion.java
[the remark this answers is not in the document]
lokka30: Treasury will eventually expand to include other APIs e.g. chat and protection, I'd rather not mess up every economy implementation by changing the EconomyAPIVersion file in any way. If it needs to be done eventually then may as well be done now. Not in this PR, I can work on it later

core/src/main/java/me/lokka30/treasury/plugin/core/utils/UpdateChecker.java
lokka30: I apologize, I was thinking of the update checker from the Spigot Wiki. All good.

lokka30 (approved):

PR #62, opened 2021-12-09
feat: transaction history

### LINKED ISSUE #33: Add Transaction History
@MrNemo64 suggested the following method: `public Collection<BankTransactionEntry> getTransactionHistory() { ... }`

Will expand on this issue later. If anyone has thoughts on this issue please drop them below :)

MrIvanPlays:
Closes #33


api/src/main/java/me/lokka30/treasury/api/economy/transaction/Transaction.java
```diff
@@ -19,6 +21,8 @@
     private final double newBalance;
     private final double transactionAmount;
     @NotNull
+    private final Currency currency;
```
MrNemo64, line 24: What do you think of saving the Currency UUID, not the Currency instance itself?

api/src/main/java/me/lokka30/treasury/api/economy/transaction/Transaction.java
```diff
@@ -27,12 +31,19 @@
...
      * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#v1_0 v1.0}
      */
-    public Transaction(final double newBalance, final double transactionAmount, @NotNull final TransactionType transactionType) {
+    public Transaction(
+            final double newBalance,
+            final double transactionAmount,
+            @NotNull final TransactionType transactionType,
+            @NotNull final Currency currency
+    ) {
         this.newBalance = AmountUtils.ensureAtLeastZero(newBalance);
```
MrNemo64, line 43: Since there is an option to support negative amounts of money this doesn't need to be at least 0 right?

MrNemo64 (commented):
So far so good but I'm failing to see how a BankAccount is able to know what player makes a transaction. We have BankTransaction that holds the player that did the transaction but how is the bank account supposed to know who made it?

Also `Transaction` feels too general of a name, could it be changet to `AccountTransaction` or `EconomyTransaction`? If we change it the `BankTransaction` name could also be changed to keep consistency with the names

MrIvanPlays:
@MrNemo64 maybe we don't need the `BankAccountTransaction` object and just have the method in the generic `Account` ? Bank accounts won't track of the player. problem solved 

200iq

MrNemo64:
> @MrNemo64 maybe we don't need the `BankAccountTransaction` object and just have the method in the generic `Account` ? Bank accounts won't track of the player. problem solved

That would be the easy way out, a posibility to have in mind. I'd like to give this some though myself and maybe others come up with something. There are other things we can work on while we think this one

lokka30:
I believe an `Optional<UUID>` should be supplied with bank account transaction methods.

* The UUID is of which player wants to execute the transaction
* It is optional because it is possible that the plugin wants to make a transaction to the bank account without a player involved, e.g. monthly tax or something
*  The UUID can then be checked by the economy provider.
  * Are they a member of the bank?
  * What permissions do they have? Can they execute the transaction they are requesting?
    * This reminds me, we probably don't have a `INSUFFICIENT_BANK_MEMBER_PERMISSIONS` constant for a failure reason

api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
```diff
@@ -152,4 +154,13 @@ public void fail(@NotNull EconomyException exception) {
...
     void deleteAccount(@NotNull EconomySubscriber<Boolean> subscription);
 
+    /**
+     * Request the {@link AccountTransaction} history of this {@code Account}
+     *
+     * @param subscription the {@link EconomySubscriber} accepting the transaction history
+     * @author MrIvanPlays
+     * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#v1_0 v1.0}
+     */
+    void retrieveTransactionHistory(@NotNull EconomySubscriber<Collection<? extends AccountTransaction>> subscription);
```
lokka30, line 164: What maximum size should the transaction history be?

Or, should this be left in the hands of the economy provider?

e.g. only supply the last 30 transactions.

api/src/main/java/me/lokka30/treasury/api/economy/account/BankAccount.java
```diff
@@ -116,4 +120,32 @@ void hasPermission(
             @NotNull EconomySubscriber<Boolean> subscription
     );
 
+    /**
+     * {@inheritDoc}
+     */
+    @Override
+    default void retrieveTransactionHistory(@NotNull EconomySubscriber<Collection<? extends AccountTransaction>> subscription) {
```
lokka30, line 127: Shouldn't we just override this method and make the subscription a `EconomySubscriber<Collection<? extends BankAccountTransaction>>`? then, remove the other method?

api/src/main/java/me/lokka30/treasury/api/economy/transaction/BankAccountTransaction.java
```diff
@@ -0,0 +1,56 @@
...
+    /**
+     * Returns the player {@link UUID}, associated with this {@code BankTransaction}
+     *
+     * @return player
+     * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#v1_0 v1.0}
+     */
+    @NotNull
+    public UUID getPlayer() {
+        return player;
+    }
```
lokka30, line 54: We should probably replace `player` with `initiator` or `bankMember`.

api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
MrIvanPlays: I will add a parameter, int lastTransactionsCount

api/src/main/java/me/lokka30/treasury/api/economy/account/BankAccount.java
MrIvanPlays: I will try doing this but I don't know if it will work because that's what i was trying to do at first but didn't work for some reason.

MrIvanPlays: not possible - we're staying as is

image

api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
lokka30: I am unsure whether 'request all transactions':
* should be possible
  a. should be done with this method, e.g. maybe `lastTransactionsCount = -1`
  b. should be done with another method
* should not be possible

api/src/main/java/me/lokka30/treasury/api/economy/account/BankAccount.java
lokka30: That sucks. Is this because the subscriber is a `Collection<? extends AccountTransaction>`?

I am unsure why we are using `<? extends ..` actually, wouldn't `Collection<AccountTransaction>` work properly?

api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
MrIvanPlays: I documented it is possible

lokka30 (approved):

api/src/main/java/me/lokka30/treasury/api/economy/account/BankAccount.java
MrIvanPlays: no because it shall also accept `BankAccountTransaction`s. Generics are stuff full of compromises and pitfalls.

lokka30:
@MrNemo64, do you approve of this PR being merged in its current state?

MrNemo64:
> I believe an `Optional<UUID>` should be supplied with bank account transaction methods.
> 
> * The UUID is of which player wants to execute the transaction
> * It is optional because it is possible that the plugin wants to make a transaction to the bank account without a player involved, e.g. monthly tax or something
> * The UUID can then be checked by the economy provider.
> * Are they a member of the bank?
> * What permissions do they have? Can they execute the transaction they are requesting?
>   
>   * This reminds me, we probably don't have a `INSUFFICIENT_BANK_MEMBER_PERMISSIONS` constant for a failure reason
> 
> This applies to _all_ bank account methods e.g. `retrieveBalance` as members require the consult permission. It also applies to the `BankAccountTransaction` class

But the methods don't include the player requesting the operation, withdraw/deposit, so how are we going to fill the optional?

Also, the Account class uses Currency for the methods, wouldn't it be better to use the UUID? For example
`void retrieveBalance(@NotNull Currency currency, @NotNull EconomySubscriber<Double> subscription);` -> `void retrieveBalance(@NotNull UUID currencyId, @NotNull EconomySubscriber<Double> subscription);`

MrIvanPlays:
Sure we will add such methods, but what we do with the other methods which are present in account?

MrNemo64:
> Sure we will add such methods, but what we do with the other methods which are present in account?

That's the main problem. If a BankAccount extends from Account then my plugin should be able to make no distinction between them. Since now the BankAccounts also need more parameters in his methods I have to worry about what account I'm using. I'm not saying that BankAccount shouldn't extend from Account, but that they work diferent enough to give us problems.

I don't really like this idea, but it came to my mind so I'm going to explain it, remember that we droped the idea of PlayerAccounts because they were going to work fundamentally the same as a BankAccount? Well since now BankAccount and Account have this problem, maybe we can revive the idea of PlayerAccounts. It would be like a BankAccount but without permissions. As I said, I really don't like this idea.
Maybe we can update the spigot page, some changes have been made and we could update the people over there and take some ideas from them.

MrIvanPlays:
essentially what we need is not an account interface, but rather player accounts and bank accounts. Or the other option: don't have banks ( the easy way out )

MrIvanPlays:
putting this on hold

MrIvanPlays:
the other way to get out of this situation is ditch `BankTransaction` entirely and do not keep track of who made the transaction (which as mrnemo mentioned we currently do not know who made the transaction at first place because the methods do not give us such information).

Jikoo:
We probably should consider adding a timestamp to transactions. That will allow us to add parameter to transaction history retrieval which will allow for pagination based on transaction age, allowing searching transaction history without retrieving all of it at once. Maybe an overload including end timestamp as well as start as a restriction? For this in particular I think implementations of Discord's message history retrieval are very instructive.

Re: transaction owner - I feel like the transaction request should contain initiator UUID no matter what the type of account is. I like the idea of an optional UUID. If its optional state is problematic it could be non-optional and Treasury could provide a default (configurable?) UUID for server transactions like tax or recurring fees. I don't see why banks should necessarily be treated differently, is there a reason to not offer regular accounts an initiator ID? I.e. an admin adjusting a balance, the server delivering currency for voting, a payment during a trade with a particular user, etc.

Regardless, we may want to add a string for a comment to make the transaction more easily user-identifiable, i.e. "Bought plot at xyz from Frank" or "City water bill" so that users know why the transaction exists if they are allowed to look at the history. ~~Can't wait for implementations using databases without PreparedStatements to get hit with SQL injection though.~~

MrIvanPlays:
I agree with Jikoo. We shall have an initiator UUID no matter what. It'll be implemented here but I will be able to work at it on Sunday.

MrNemo64:
> We probably should consider adding a timestamp to transactions. That will allow us to add parameter to transaction history retrieval which will allow for pagination based on transaction age, allowing searching transaction history without retrieving all of it at once. Maybe an overload including end timestamp as well as start as a restriction? For this in particular I think implementations of Discord's message history retrieval are very instructive.

I wanted to say this, but always forgot


> Re: transaction owner - I feel like the transaction request should contain initiator UUID no matter what the type of account is. I like the idea of an optional UUID. If its optional state is problematic it could be non-optional and Treasury could provide a default (configurable?) UUID for server transactions like tax or recurring fees. I don't see why banks should necessarily be treated differently, is there a reason to not offer regular accounts an initiator ID? I.e. an admin adjusting a balance, the server delivering currency for voting, a payment during a trade with a particular user, etc.

I hadn't thought of it that way, and I like the idea


> Regardless, we may want to add a string for a comment to make the transaction more easily user-identifiable, i.e. "Bought plot at xyz from Frank" or "City water bill" so that users know why the transaction exists if they are allowed to look at the history.

I support this idea. We can add an `Optional<String> getMessage()` without too much trouble. Can can overload the methods from Account to make it easyer, for example
```Java
void setBalance(double amount, UUID currencyId, UUID initiatorUUID, String message, EconomySubscriber<Double> subscription);

default void setBalance(double amount, UUID currencyId, String message, EconomySubscriber<Double> subscription) {
  setBalance(amount, currencyId, null, message, subscription);
}

default void setBalance(double amount, UUID currencyId, UUID initiatorId, EconomySubscriber<Double> subscription) {
  setBalance(amount, currencyId, initiatorId, null, subscription);
}

default void setBalance(double amount, UUID currencyId, EconomySubscriber<Double> subscription) {
  setBalance(amount, currencyId, null, null, subscription);
}
```
Although I know @lokka30 doesn't like bloating apis

> ~Can't wait for implementations using databases without PreparedStatements to get hit with SQL injection though.~

~oh boy~

lokka30:
I like all the ideas Jikoo proposed.

MrIvanPlays:
I think we should also return to the name `Transaction` for the `AccountTransaction` class as we are not going to have `BankAccountTransaction` anymore.

MrNemo64:
> I think we should also return to the name `Transaction` for the `AccountTransaction` class as we are not going to have `BankAccountTransaction` anymore. 

I don't think so, a transaction can be a lot of things, for example a LandTransaction, an EconomyTransaction... By naming the economy transactions just "Transaction" we may create confusion

MrIvanPlays:
then im naming it `EconomyTransaction`

api/src/main/java/me/lokka30/treasury/api/economy/transaction/EconomyTransaction.java
```diff
@@ -0,0 +1,227 @@
...
+     * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#v1_0 v1.0}
+     */
+    public EconomyTransaction(
+            @NotNull UUID currencyUUID,
+            @NotNull UUID initiatorUUID,
+            @NotNull EconomyTransactionType economyTransactionType,
+            final double transactionAmount
+    ) {
+        this.currencyUUID = Objects.requireNonNull(currencyUUID, "currencyUUID");
+        this.initiatorUUID = Objects.requireNonNull(initiatorUUID, "initiatorUUID");
```
MrNemo64, line 58: What if the initiator is, for example, a vote plugin? Do we have to give an UUID to the vote plugin?

api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
```diff
@@ -162,4 +205,51 @@ public void fail(@NotNull EconomyException exception) {
...
+     * really need all the transactions, please use this with care, because an economy provider might do a query to a database
+     * .</i>
+     *
+     * @param transactionCount the count of the transactions wanted
+     * @param subscription     the {@link EconomySubscriber} accepting the transaction history
+     * @author MrIvanPlays
+     * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#v1_0 v1.0}
+     */
+    void retrieveTransactionHistory(
+            int transactionCount, @NotNull EconomySubscriber<Collection<EconomyTransaction>> subscription
```
MrNemo64, line 223: Maybe put each parameter in a diferent line?

api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
```diff
@@ -162,4 +205,51 @@ public void fail(@NotNull EconomyException exception) {
...
+    /**
+     * Request the {@link EconomyTransaction} history, starting from the {@link Temporal} {@code from} specified, of this
+     * {@code Account}.
+     *
+     * @param from         the timestamp to get the transactions from
+     * @param subscription the {@link EconomySubscriber} accepting the transaction history
+     * @author MrIvanPlays
+     * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#v1_0 v1.0}
+     */
+    default void retrieveTransactionHistory(
```
MrNemo64, line 235: Since we are having the option to specify a `from` could we also specify a `from` and a `to` in another method?

api/src/main/java/me/lokka30/treasury/api/economy/transaction/EconomyTransaction.java
```diff
@@ -0,0 +1,227 @@
...
+    /**
+     * Creates a new account transaction object.
+     *
+     * @param currencyUUID           the currency's UUID the transaction was made into
+     * @param initiatorUUID          the one who initiated the transaction
+     * @param economyTransactionType the transaction type
+     * @param transactionAmount      the amount which to deposit/withdraw
+     * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#v1_0 v1.0}
+     */
+    public EconomyTransaction(
```
MrNemo64, line 51: We could add an optional string representing a message for the transaction

api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
MrIvanPlays: the editorconfig does these changes. that's really not necessary

api/src/main/java/me/lokka30/treasury/api/economy/transaction/EconomyTransaction.java
MrIvanPlays: ill make it nullable

MrNemo64: Not necessarily make it nullable
Maybe this idea is a bit over the top but, what if we have an interface `TransactionInitiator` and have several subclasses/subinterfaces like `PlayerTransactionInitiator`, `PluginTransactionInitiator` and `ServerTransactionInitiator` each one of these holding the information about the initiatior. The `PlayerTransactionInitiator` and UUID, the `PluginTransactionInitiator` the plugin name etc

api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
MrNemo64: I thought it would split those line by default

MrIvanPlays: well the editorconfig is something which isn't ideal. it sometimes does and sometimes it tells such splits are "custom line splits". doesn't really matter tho as it is still readable like this.

api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
MrIvanPlays: done

api/src/main/java/me/lokka30/treasury/api/economy/transaction/EconomyTransaction.java
MrIvanPlays: done

api/src/main/java/me/lokka30/treasury/api/economy/transaction/EconomyTransactionInitiator.java
```diff
@@ -0,0 +1,117 @@
...
+     */
+    @NotNull
+    static <T> EconomyTransactionInitiator<T> createInitiator(@NotNull Type type, T data) {
+        Objects.requireNonNull(type, "type");
+        switch (type) {
+            case PLAYER:
+                if (data == null) {
+                    throw new IllegalArgumentException("Player initiator with no data");
+                }
+                if (!UUID.class.isAssignableFrom(data.getClass())) {
```
MrNemo64, line 63: Why not `data instanceof UUID`?

api/src/main/java/me/lokka30/treasury/api/economy/transaction/EconomyTransactionInitiator.java
```diff
@@ -0,0 +1,117 @@
...
+                    throw new IllegalArgumentException("Player initiator with data which is not an UUID");
+                }
+                break;
+            case PLUGIN:
+                if (data == null) {
+                    throw new IllegalArgumentException("Plugin initiator with no data");
+                }
+                if (!String.class.isAssignableFrom(data.getClass())) {
```
MrNemo64, line 71: Why not `data instanceof String`?

api/src/main/java/me/lokka30/treasury/api/economy/transaction/EconomyTransactionInitiator.java
```diff
@@ -0,0 +1,117 @@
...
+ * Represents an initiator of {@link EconomyTransaction}.
+ * <p>Initiator is something/someone, triggering a transaction.
+ * <p>Initiators are divided in 3 types: {@link Type#PLAYER}, {@link Type#PLUGIN} and {@link Type#SERVER}.
+ * The PLAYER and PLUGIN types should always provide data which matches types:
+ * <ul>
+ *     <li>For PLAYER type: data type should be an {@link UUID} of the player, whom initiated the transaction</li>
+ *     <li>For PLUGIN type: data type should be am {@link String}, fulfilled with the plugin name, whom initiated the
+ *     transaction.</li>
+ * </ul>
+ * <p>It is not necessary to provide data for type SERVER.
```
MrNemo64, line 17: Since server has always the same data (no data) it could just be an instance 
```Java
  private static final EconomyTransactionInitiator<?> SERVER_INITIATOR = ...

  static EconomyTransactionInitiator<?> getServerInitiator() {
    return SERVER_INITIATOR;
  }
```
And then we could have separate static methods to create player and plugin initiators, making the comparison to see if it's the correct data easyer

MrIvanPlays: > And then we could have separate static methods to create player and plugin initiators, making the comparison to see if it's the correct data ~~easyer~~ easier

Bloat has gone too much.

api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
```diff
@@ -77,46 +79,109 @@ void setBalance(
...
+     * @param subscription the {@link EconomySubscriber} accepting the new balance
+     * @author MrIvanPlays
+     * @see Account#setBalance(double, EconomyTransactionInitiator, Currency, EconomySubscriber)
+     * @see Account#doTransaction(EconomyTransaction, EconomySubscriber)
+     * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#v1_0 v1.0}
+     */
+    default void withdrawBalance(
+            double amount,
+            @NotNull EconomyTransactionInitiator<?> initiator,
+            @NotNull Currency currency,
```
MrNemo64, line 118: I have commented this several times but, wouldn't it be better to have as parameter the UUID of the currency instead of the instance itself?

MrIvanPlays:
```java
withdrawBalance(20, initiator, currency, // ...
// vs
withdrawBalance(20, initiator, currency.getCurrencyId(), // ...
```
BUT WAIT THERE'S MORE TO THAT!!!! Every time the providing plugin needs a Currency instance it would have to query it from the CurrencyManager, whilst for the depending plugin it is easier, because it can store the `Currency` object somewhere static

To answer your question, no.

api/src/main/java/me/lokka30/treasury/api/economy/transaction/EconomyTransactionInitiator.java
MrNemo64: > Bloat has gone too much.

Alright. But what about the server initiator? `createInitiator` could return that instance

MrIvanPlays: We have
```java
Collections.emptyList();
```
but 
```java
new ArrayList<>();
```
also returns an empty list if not modified, doesn't it?

MrNemo64: I don't follow you, what are you trying to say?

MrIvanPlays: > I don't follow you, what are you trying to say?

~~learn english~~ I'm going to try to explain in code

```java
Collections.emptyList(); // isEmpty() returns true
new ArrayList<>(); // at the point of creation isEmpty() returns true
```

What does this mean? It means that you can also create an empty list even though `Collections.emptyList()` exists.

api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
MrNemo64: I forgot that CurrencyManager returns an `Optional<Currency>`, I thought we would have to use a subscriber to get a currency so it would get messy fast. 
Now that I've checked the currency manager, there is really no diference between them so I don't mind either option

api/src/main/java/me/lokka30/treasury/api/economy/transaction/EconomyTransactionInitiator.java
MrNemo64: What I'm trying to say is, why create and instance of a basically inmutable object every time is needed. Look for example at the [Optinal](https://github.com/openjdk/jdk/blob/fe2ae8e38bd8660cc637f13f425ccea17a11aa4f/src/java.base/share/classes/java/util/Optional.java) class

```Java
    private static final Optional<?> EMPTY = new Optional<>(null);

    public static <T> Optional<T> ofNullable(T value) {
        return value == null ? (Optional<T>) EMPTY
                             : new Optional<>(value);
    }

    public static<T> Optional<T> empty() {
        @SuppressWarnings("unchecked")
        Optional<T> t = (Optional<T>) EMPTY;
        return t;
    }
```
There is only one instance of an empty optional because it's basically an inmutable object, why waste time creating instances of it?

MrIvanPlays: instantiating an object in java is literally if not done in nano, then done in pico seconds. literally unmeasurable "pressure" if we count it as such. The `Optional` class is a completely different story.

lokka30:
So, how are these transactions created in the first place? Are we going to modify the deposit, set and withdraw methods to be supply enough data to fill the transaction's information? Or, should we just replace the the three methods with `sendTransaction`?

MrIvanPlays:
We have deposit and withdraw methods with which you are able to fulfill a transaction object which are implemented as default methods and then you have doTransaction which accepts EconomyTransaction and a EconomySubscriber. The transaction class has a builder for easier creation of such objects. Could've just looked at the code and I wouldn't need to explain this to ya @lokka30

api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
lokka30: Is this conversation resolved?

MrIvanPlays: Yes.

lokka30 (approved):

MrNemo64 (commented):
If I create a transaction and then use the `doTransaction` twice, then I have done 2 transactions in the exact same moment as the instant of the transaction is not updated, even if there is hours between each call. Should we do something about that? mention it in the coments or make `doTransaction` accept an Instant of time?

MrIvanPlays:
I will manually recreate the commit and push to master after jikoo's approval since branch got fucked up. Exact diff from master is this commit: https://github.com/lokka30/Treasury/pull/62/commits/575e7b1771779a27f8c91bc57baf0aa6dd507f3d

MrNemo64:
There is no way to undo the wrong commits?

MrIvanPlays:
unfortunately no since its a force push. I will cherry pick the commit from my branch I think it would work.

api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
```diff
@@ -152,4 +263,91 @@ public void fail(@NotNull EconomyException exception) {
...
+     * <p>If the specified {@code transactionCount} is {@code -1}, then this will return all the transactions. <i>If you
+     * really need all the transactions, please use this with care, because an economy provider might do a query to a database
+     * .</i>
+     *
+     * @param transactionCount the count of the transactions wanted
+     * @param subscription     the {@link EconomySubscriber} accepting the transaction history
+     * @author MrIvanPlays
+     * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#v1_0 v1.0}
+     */
+    void retrieveTransactionHistory(
```
Jikoo, line 289: I really don't like the idea of potentially thousands of transactions being loaded, even if it isn't done on the main thread. Were I to implement an economy plugin I might offer a Stream of transactions that would request additional transactions as necessary to complete its history, but I would not ever support loading all transactions in one single bucket under any circumstances - I'd cap this somewhere in the 25-100 range and force you to use the timestamp-based options if you wanted more.

I would vastly prefer to see an overload `retrieveTransactionHistory(int transactionCount, @NotNull Temporal from, @NotNull Temporal to, @NotNull EconomySubscriber<Collection<EconomyTransaction>> subscription)`.
This would then become a default implementation:
```java
default void retrieveTransactionHistory(
    int transactionCount, @NotNull EconomySubscriber<Collection<EconomyTransaction>> subscription
) {
  retrieveTransactionHistory(transactionCount, NOW, EPOCH_START, subscription);
}
```

api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
```diff
@@ -152,4 +263,91 @@ public void fail(@NotNull EconomyException exception) {
...
+        } else {
+            fromInstant = Instant.from(from);
+        }
+        Instant toInstant;
+        if (to instanceof Instant) {
+            toInstant = (Instant) to;
+        } else {
+            toInstant = Instant.from(to);
+        }
+        retrieveTransactionHistory(-1, new EconomySubscriber<Collection<EconomyTransaction>>() {
```
Jikoo, line 337: Retrieving the entirety of transaction history and then filtering it is horrendously inefficient, it defeats the primary purpose of loading only a segment of the history. The implementation should handle this method in its entirety so that unnecessary transaction entries are not loaded at all.

Jikoo (changes requested):

MrIvanPlays:
@Jikoo I removed the (from, EconomySubscriber) method. What do you think of the implementation now?

Jikoo (approved):
LGTM 👍

MrIvanPlays:
https://github.com/lokka30/Treasury/commit/25f9307867bc8af8dae391d064832e349d3e4bed

PR #68, opened 2021-12-11
Add method to get all accounts a player is member of/has X permission on

### LINKED ISSUE #67: Add method to get all accounts a player is member of/has X permission on
Right now if I want to get all the accounts a player has, for example, `withdraw` permission on then I have to go thruough all accounts and check, not really optimal. The `EconomyProvider` could have a 
`void retriveAllAccountsPlayerIsMemberOf(UUID player, EconomySubscriber<Collection<? extends Acount>>)`
`void retriveAllAccountsIdPlayerIsMemberOf(UUID player, EconomySubscriber<Collection<UUID>>)`
`void retriveAllAccountsPlayerHasPermission(UUID player, BankAccountPermission[] permissions, EconomySubscriber<Collection<? extends BankAccount>>)`
`void retriveAllAccountsIdPlayerHasPermission(UUID player, BankAccountPermission[] permissions, EconomySubscriber<Collection<UUID>>)`

MrNemo64:
Intends to close #67

### DIFF
--- a/api/src/main/java/me/lokka30/treasury/api/economy/EconomyProvider.java
+++ b/api/src/main/java/me/lokka30/treasury/api/economy/EconomyProvider.java
@@ -4,9 +4,7 @@
 
 package me.lokka30.treasury.api.economy;
 
-import java.util.ArrayList;
 import java.util.Collection;
-import java.util.List;
 import java.util.Set;
 import java.util.UUID;
 import me.lokka30.treasury.api.economy.account.Account;
@@ -16,7 +14,6 @@ import me.lokka30.treasury.api.economy.account.PlayerAccount;
 import me.lokka30.treasury.api.economy.currency.Currency;
 import me.lokka30.treasury.api.economy.misc.EconomyAPIVersion;
 import me.lokka30.treasury.api.economy.misc.OptionalEconomyApiFeature;
-import me.lokka30.treasury.api.economy.response.EconomyException;
 import me.lokka30.treasury.api.economy.response.EconomySubscriber;
 import org.jetbrains.annotations.NotNull;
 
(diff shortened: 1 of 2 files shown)


MrNemo64:
Should we keep the account and id methods or just one of them?

lokka30:
> Should we keep the account and id methods or just one of them?

I think just the ID methods. There is extra overhead by requesting multiple bank accounts from the provider.

lokka30 (changes requested):
I think them method should be renamed to `retrieveAllBankAccountsMemberOf`, even with the permission argument.

MrNemo64:
> I think them method should be renamed to `retrieveAllBankAccountsMemberOf`, even with the permission argument.

`retrieveAllBankAccountsPlayerHasPermission` maybe should also return the personal account of the player? After all, tecnically the player has all permissions on that account, same with `retrieveAllAccountsPlayerIsMemberOf` a player isalways a member of his personal account
or do we want these methods to just return the bank accounts?

lokka30:
> > I think them method should be renamed to `retrieveAllBankAccountsMemberOf`, even with the permission argument.
> 
> `retrieveAllBankAccountsPlayerHasPermission` maybe should also return the personal account of the player? After all, tecnically the player has all permissions on that account, same with `retrieveAllAccountsPlayerIsMemberOf` a player isalways a member of his personal account or do we want these methods to just return the bank accounts?

Just the bank accounts. That's all people using that method will be looking for. It's a given that the player can access their own account :)

MrNemo64:
Should we add the `@see` tag on methods like `retrieveBankAccountIds`?

api/src/main/java/me/lokka30/treasury/api/economy/EconomyProvider.java
```diff
@@ -140,6 +142,36 @@
...
+     * Request all the {@link BankAccount bank accounts} where the given player has the given permissions.
+     *
+     * @param playerId     the player
+     * @param permissions  the permissions that the given player has to have on the {@link BankAccount account}
+     * @param subscription the {@link EconomySubscriber} accepting the resulting value
+     * @author MrNemo64
+     * @see #retrieveAllBankAccountsPlayerIsMemberOf(UUID, EconomySubscriber)
+     * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#v1_0 v1.0}
+     */
+    void retrieveAllBankAccountsPlayerHasPermission(
```
MrIvanPlays, line 169: In my opinion this should be UUID, EconomySubscriber, BankAccountPermission... because it allows you to do stuff as

```java
economyProvider.retrieveAllBankAccountsPlayerHasPermission(playerId, subscription, BankAccountPermision.FOO, BankAccountPermission.BAR);
```

MrIvanPlays (approved):
1 thing I have a comment on which is more of a nitpick, but a QOL change.

api/src/main/java/me/lokka30/treasury/api/economy/EconomyProvider.java
MrNemo64: I thought about that but then the arguments get splitted into multiple lines. TBH I also prefer this way of doing it, if @Jikoo and @lokka30 can give their opinions

Jikoo: I do also prefer varargs for variable length array arguments.

lokka30 (approved):

lokka30:
Thanks! Ready to merge when you are.

lokka30:
@Jikoo do you approve of this PR so far?

api/src/main/java/me/lokka30/treasury/api/economy/EconomyProvider.java
```diff
@@ -140,15 +145,162 @@
...
+                                    subscription.fail((EconomyException) throwable);
+                                } else {
+                                    subscription.fail(new EconomyException(FailureReason.OTHER_FAILURE, throwable));
+                                }
+                                return null;
+                            }).thenAccept(val -> {
+                        if (!val) {
+                            modified.set(true);
+                        }
+                    });
```
MrNemo64, line 256: I don't think we should check if the player is a memeber on this method, only if they have the permissions
A player will be a bank memeber as long as he has at least one permission on it so if they have the permissions they are bank members

MrNemo64 (commented):
Maybe it's worth considering changing `BankAccount#hasPermission(UUID, BankAccountPermission, EconomySubscriber)` to `BankAccount#hasPermission(UUID, EconomySubscriber, BankAccountPermission...)` after seeing this

Jikoo (approved):
LGTM 👍

MrIvanPlays:
Thanks!

PR #71, opened 2021-12-14
Rework bank accounts

### LINKED ISSUE #58: Remove bank owners, use just members instead
@MrNemo64, what do you think about this idea? I believe there should only be bank members, but of course each bank member can have different permissions.

### LINKED ISSUE #70: Add bank account names
I've had this idea for quite some time now and I can't figure out if it's worth it to add to the API. My thought process what that with code we diferenciate accounts by their uuid, in game there is an easy diferenciation between player and bank accounts but what about diferenciating several bank accounts? If a player owns 2 diferent bank accounts, how is he supposed to diferenciate them? I don't thing the uuid will be usefull for them. So I thought of adding a `String name` field to bank accounts, so you can diferenciate them ingame but it feels like it should be left for any plugin that needs a name for a bank account, this however brings its problems.
I don't want to bloat the API with unecessary features and I don't think this is really necessary but it has been on my mind for quite some time so I just wrote it.

Just to clarify, if we add this there should be no api provided way of getting bank accounts by the name.

To give an example: My plugin wants to charge X coins on a player and I show in chat/open a menu with all the accounts that they can withdraw from. The player could just select the account based on the income but what if they want to withdraw from an especific account? Do that have to keep track of the coins in that account to identify it?
Seems like this is a my plugin problem and I should be the one to solve it, but if we have 2+ plugins that doing this then having a unique name for every bank account avoids the pain for the player of naming all bank accounts several times and unnecessary repetition of data between plugins

I want to hear your thoughts on this

MrIvanPlays:
Closes #58
Closes #70

### DIFF
--- a/api/src/main/java/me/lokka30/treasury/api/economy/EconomyProvider.java
+++ b/api/src/main/java/me/lokka30/treasury/api/economy/EconomyProvider.java
@@ -20,6 +20,7 @@ import me.lokka30.treasury.api.economy.response.EconomyException;
 import me.lokka30.treasury.api.economy.response.EconomySubscriber;
 import me.lokka30.treasury.api.economy.response.FailureReason;
 import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
 
 /**
  * Implementors providing and managing economy data create a class
@@ -133,7 +134,20 @@ public interface EconomyProvider {
      * @author Jikoo
      * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#v1_0 v1.0}
      */
-    void createBankAccount(@NotNull UUID accountId, @NotNull EconomySubscriber<BankAccount> subscription);
+    default void createBankAccount(@NotNull UUID accountId, @NotNull EconomySubscriber<BankAccount> subscription) {
+        createBankAccount(null, accountId, subscription);
+    }
+
+    /**
+     * Request the creation of a {@link BankAccount} for a {@link UUID} {@code accountId} with {@link String} {@code name}.
+     *
+     * @param name         the name of the account
+     * @param accountId    the {@code UUID} of the account
+     * @param subscription the {@link EconomySubscriber} accepting the resulting value
+     * @author MrIvanPlays, Jikoo
+     * @since {@link EconomyAPIVersion#v1_0 v1.0}
+     */
+    void createBankAccount(@Nullable String name, @NotNull UUID accountId, @NotNull EconomySubscriber<BankAccount> subscription);
 
     /**
      * Request all {@link UUID UUIDs} with associated {@link BankAccount BankAccounts}.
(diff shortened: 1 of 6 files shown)


MrIvanPlays:
@MrNemo64

MrNemo64:
As said [here](https://github.com/lokka30/Treasury/issues/58#issuecomment-990819182), should we have an ` addMember` method? I suppose it can 
just return false if needs a permission.

It could be worth mentioning in the docs that "a member is usually any player with at least one permission on the account unless the provider allows for members without permissions"

Also I may be blind but are there methods to add and remove permissions from a player?

MrIvanPlays:
You are blind. There is an `addMember` method. But as of right now it does not accept permission(s).

MrIvanPlays:
@MrNemo64 do you agree with the proposed solution?

MrNemo64:
> @MrNemo64 do you agree with the proposed solution? 

So if I set `CONSULT` to true, the player can consult the account
If I set `CONSULT` to unspecified the player is not explicitly allowed to consult the balance but we also won't block his consult requests. It's like the default one
If I set `CONSULT` to false, there is no way the player can consult the account and all his requests are blocked
`retrieveAllBankAccountsPlayerHasPermission` would return all bank accounts a player has set the permission with true but not with unspecified or false

If i got it correctly then I kinda like this idea more than my original one, but I think we should try to simply the description
I'd like to hear other opinions as unspecified may create problems if a consumer plugin doesn't check permissions

I feel like TriState should be moved to a misc package out of the economy api as it's not economy specific

Can we add the member definition to all methods that do something with the members? 

Right now I can't review the changes as much as I'd like, Tuesday probably I'll be able to do a full review

MrIvanPlays:
Let me explain this quick ; 

if you set `CONSULT` to `UNSPECIFIED` and `CONSULT` is the only permission they have, they will be removed from the bank. 

if you set `CONSULT` to `TRUE` and they didn't have any permissions they become a member of the bank with the ability to consult it.

if you set `CONSULT` to `FALSE` and they have other permissions, they won't be able to consult the bank, yet they are still bank members. if you set `CONSULT` to `FALSE` and they didn't have any permissions, they shall not be considered bank members, although here we can open a bracket that different economy providers can decide if they want to allow such behavior.

> Can we add the member definition to all methods that do something with the members?

I will add it to `setPermission` and `hasPermission` too.

> also, are we going to allow to change the name of an account? 

My initial thought was the name not to be changeable, but since you bring that I will add such method too.

MrIvanPlays:
looks like wherever I decide to touch it needs a rework :eyes: 

I hope bank accounts are going to be the final thing we're going to change. ( literally there's nothing else to change)

lokka30 (approved):
Looks good to me, no feedback other than great work! Thanks :)

Jikoo:
Looks good to me! Sorry I was slow on the response to this one.

Jikoo (approved):

Issue #74, opened 2021-12-18
Currency System Overhaul


creatorfromhell:
The currency system as it stands is very restrictive towards economy plugins who generally have their own currency system in place. While it is very nicely designed, it's designed as a system that should be inside an economy plugin itself and not inside of an API.

In general, the currency API oversteps what an API should provide.

Oversteps:
- Forcing the currency manager inside Treasury onto economy plugins
- Forcing the conversion standards inside Treasury onto economy plugins.
   - This really should just be a method that .to(Currency other) that is implemented by economy plugins
- Forcing currency UUIDs that are used nowhere inside the API onto economy plugins.  
- Forcing a default formatting system onto economy plugins.
- Forcing the currency registration system onto economy plugins.

As I stated before, in general the system is great, but it is a system that shouldn't be inside an API, but rather an economy plugin that implements the API.

Why is this a bad thing?
  The reason overstepping is a bad thing is it reduces the likelihood of adoption.

MrNemo64:
I agree with most points here but

> Forcing currency UUIDs that are used nowhere inside the API onto economy plugins.

Since the methods take the Currency instance itself and not the UUID we could tecnically remove them but how would a plugin request the instance of the currency?

> Forcing a default formatting system onto economy plugins.

We need to provide a formatting system for plugins to use, how else are they supposed to display a currency?

> Forcing the currency registration system onto economy plugins.

Now that you mention it, i don't know if its good or not to have it

creatorfromhell:
> I agree with most points here but
> 
> > Forcing currency UUIDs that are used nowhere inside the API onto economy plugins.
> 
> Since the methods take the Currency instance itself and not the UUID we could tecnically remove them but how would a plugin request the instance of the currency?


They would still use a currency identifier, but the issue is that forcing an economy plugin to specify it to be a UUID is rather radical. A string identifier is suitable enough, which provides freedom to be an UUID or not.

> 
> > Forcing a default formatting system onto economy plugins.
> 
> We need to provide a formatting system for plugins to use, how else are they supposed to display a currency?

When I meant "forcing a formatting system" I meant that forcing them to utilize Locale, when most plugins don't even have a Locale-based translation system in them is pretty restrictive. Simply having a formatting method with an amount and precision suffices, It should be up to the implementation, in my opinion.

> 
> > Forcing the currency registration system onto economy plugins.
> 
> Now that you mention it, i don't know if its good or not to have it

[...]

Issue #44, opened 2021-10-26
Split EconomyProvider interface or remove hasXSupport methods


Jikoo:
Edit:
After the discussion here, I feel like the `hasXSupport` methods are a bad idea and I would actually like to see Treasury drop the idea of not implementing the entirety of its interfaces.

> It really doesn't make sense to me to have 50% of the methods be designed to not be implemented. Anyone who does offer Treasury support should be able to offer bank support with minimal extra work due to how similar the accounts are.

If we do truly want a method to check for support that should likely just be rolled into the API version system, not the provider interface(s). The javadocs for each method state the API version it was added in and the provider states the API version it was programmed against.

Original writeup:
I would like to see `EconomyProvider` separated into several interfaces.

As a developer, it makes it a lot easier to work with player/bank accounts if I know exactly what is available to me rather than constantly re-checking if banks are supported and whatnot.

As a server owner, the increased granularity makes it possible to invent a solution that more exactly suits my needs by selecting the exact bank and player account implementations I want.

* `TreasuryApiProvider`
  * Provide Treasury API version implemented, owning plugin
* ~~`CurrencyProvider extends TreasuryApiProvider`~~
  * ~~Provides currency definitions and handles parsing~~
  * ~~Should define currency management (creation, deletion, modification, etc.)~~
  * ~~Treasury should probably implement and provide at `ServicePriority.Lowest`~~
    * See #40
  * Definitely needs improvement, but outside the scope of this issue
* `AccountProvider extends TreasuryApiProvider`
  * Provides a `requestCurrenciesInUse` or similar method
    * Allows `CurrencyProvider` to support undefined currencies, making migration easier.
* `PlayerAccountProvider extends AccountProvider`
  * Provides `PlayerAccounts`
* `BankAccountProvider extends AccountProvider`
  * Provides `BankAccounts`
  * Methods should not clash with `PlayerAccountProvider`
    * Makes it possible to create a generic `MyEconomyImpl implements PlayerAccountProvider, BankAccountProvider` to mimic Vault style interface usage

MrIvanPlays:
I mean we can clean it up after these calls. Its just an idea I am not forcing it to go into the implementation.

Jikoo:
I've edited OP to reflect my current stance on this for clarity.

> An idea I've just got in my mind. Whenever a economy provider tries to register into treasury, make a test of all features. If it fails at any feature, do not register it.

[...]

PR #75, opened 2021-12-18
Currency System Rework. Addresses #74.

creatorfromhell:
Addresses #74 https://github.com/lokka30/Treasury/issues/74


MrNemo64 (changes requested):
I don't like the idea of removing the CurrencyManager and moving the currency methods to the EconomyProvider, we also lose the hability to parse any string into a curremcy and an amount

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
```diff
@@ -1,236 +1,141 @@
...
+    /**
+     * Used to translate an amount to a user readable format with the specified amount of decimal places.
+     *
+     * @param amount The amount to format.
+     * @param precision The amount of decimal digits to use when formatting.
+     * @return The formatted text.
+     * @author creatorfromhell
+     * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#v1_0 v1.0}
+     */
+    String format(@NotNull Double amount, @NotNull Integer precision);
```
MrNemo64, line 140: `@NotNull Double` why not just a `double`?
`@NotNull Integer` why not just a `int`?

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
```diff
@@ -1,236 +1,141 @@
...
+     * @return The formatted text.
+     * @author creatorfromhell
      * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#v1_0 v1.0}
      */
-    @NotNull
-    public String formatBalance(double amount, @NotNull Locale locale) {
-        Objects.requireNonNull(locale, "locale");
-        return balanceFormatter.apply(amount, locale);
-    }
+    String format(@NotNull Double amount);
```
MrNemo64, line 129: `@NotNull Double` why not just a `double`?

core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/economy/migrate/MigrationEconomy.java
```diff
@@ -32,7 +33,82 @@
...
+            @Override
+            public void deformat(@NotNull final String formatted, @NotNull final EconomySubscriber<Double> subscription) {
+                try {
+                    Double value = Double.parseDouble(formatted);
+                    subscription.succeed(value);
+                } catch(Exception ignore) {
+                    subscription.fail(new EconomyException(FailureReason.OTHER_FAILURE, "Formatted is not a valid Double value" +
+                            "."));
+                }
+            }
```
MrNemo64, line 95: This should always fail

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
creatorfromhell: This is based on the fact that Treasury utilizes @NotNull when primitive types cannot be null nor can they be compared to null in java.

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
creatorfromhell: This is based on the fact that Treasury utilizes @NotNull when primitive types cannot be null nor can they be compared to null in java.

core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/economy/migrate/MigrationEconomy.java
creatorfromhell: > I don't like the idea of removing the CurrencyManager and moving the currency methods to the EconomyProvider, we also lose the hability to parse any string into a curremcy and an amount

Could you elaborate on this? Also forcing that currency manager isn't a good practice as it has no business being in an API. An API should simply provide the connection for the actions to happen, not do the action for them.

creatorfromhell: > This should always fail

Will fix this.

MrNemo64: > Could you elaborate on this

Ideally a class should only be in charge of 1 things. EconomyProvider is already in charge of Accounts and the CurrencyManager handled the currencies, merging them gives the EconomyProvider all the responsability.

We discussed this here anyways
https://github.com/lokka30/Treasury/issues/44
i suggested this but it seems like it went unnoticed
> About splitting the provider into several providers, could we have all providers splited as described by Jikoo or something like that and then have the economy provider extend all those interfaces?
This way if you want to use the separated parts them you just cast, if you want all use the EconomyProvider. This would allow also for the implementation to have its logic separated _by default_

However a EconomyProvider implementation could just delegate the tasks to other objects, kinda like the Bukkit class

So it comes down to my preferences

creatorfromhell: > > Could you elaborate on this
> 
> Ideally a class should only be in charge of 1 things. EconomyProvider is already in charge of Accounts and the CurrencyManager handled the currencies, merging them gives the EconomyProvider all the responsability.
> 
> We discussed this here anyways #44 i suggested this but it seems like it went unnoticed
> 

Except for the fact that EconomyProvider still provided methods for the primary currency. This goes against the very essence of the class's documented description which is for "managing economy data" as outlined in the javadocs of the class. In essence if it should simply be for account management it shouldn't be called an EconomyProvider and shouldn't have the methods for supportedAPIVersion or supportedFeatures. Hence why CurrencyManager makes no sense to have when currencies are data for economies.

MrNemo64: You're right there

creatorfromhell: > You're right there

You're not wrong though it really comes down to preference, which isn't really where an API should come into place.

lokka30 (changes requested):

MrIvanPlays (changes requested):
The whole idea of the current implementation is for ease of currency creation not just by the economy provider, but by plugins using the API. 

> Forcing a default formatting system onto economy plugins.

Nope. Currently the `Currency` class accepts a BiFunction<Double, Locale, String> for formatting a currency. 

> Forcing the conversion standards inside Treasury onto economy plugins.

We'd be better off changing the conversion coefficient to a BiFunction<Double, Currency, Double> for independent conversion rather than nuking everything and redoing it in the way we wanted to avoid.

> Forcing the currency manager inside Treasury onto economy plugins

Bukkit is also forcing the plugin manager onto plugins. Bukkit is also forcing you to create plugin.yml so your plugin runs. Bukkit is also forcing you to implement `CommandExecutor` in order to create a  command and it is also forcing you to define it in the plugin.yml. Bukkit is forcing you to implement `Listener` so you can register event handlers.

> Forcing currency UUIDs that are used nowhere inside the API onto economy plugins.

Just because they are not used in the treasury plugin does not mean they won't be used _at all_. Currency UUIDs are a way to identify a currency, and a way to store more data with less memory in a set, list, etc.

creatorfromhell:
> The whole idea of the current implementation is for ease of currency creation not just by the economy provider, but by plugins using the API.

The new system is still easy to create new currencies outside the provider. You just call .registerCurrency.
> 
> > Forcing the conversion standards inside Treasury onto economy plugins.
> 
> We'd be better off changing the conversion coefficient to a BiFunction<Double, Currency, Double> for independent conversion rather than nuking everything and redoing it in the way we wanted to avoid.

Again it's not an API's place to dictate how something should be handled. While I agree with the fact there should be a method for conversion, it shouldn't be dictated how it's achieved, or the system that should be used. 

> 
> > Forcing the currency manager inside Treasury onto economy plugins
> 
> Bukkit is also forcing the plugin manager onto plugins. Bukkit is also forcing you to create plugin.yml so your plugin runs. Bukkit is also forcing you to implement `CommandExecutor` in order to create a command and it is also forcing you to define it in the plugin.yml. Bukkit is forcing you to implement `Listener` so you can register event handlers.

That kinda hit the marker a little with comparisons. This would be essentially like registering a permission node to a player without doing so in a permissions plugin. It's an extreme overreach.

> 
> > Forcing currency UUIDs that are used nowhere inside the API onto economy plugins.
> 
> Just because they are not used in the treasury plugin does not mean they won't be used _at all_. Currency UUIDs are a way to identify a currency, and a way to store more data with less memory in a set, list, etc.

Yes but it should be up to implementations.

MrIvanPlays:
> it shouldn't be dictated how it's achieved, or the system that should be used.

Looks like you don't really understand how github reviewing works. I'm suggesting a way to implement it which for me looks more viable and better with the current implementation. If you have a better method please show it. I don't like the thing that I need to implement 1000s of interfaces in order to make an economy provider with 1 currency work.

> This would be essentially like registering a permission node to a player without doing so in a permissions plugin. It's an extreme overreach.

I don't understand your point here. 

> Yes but it should be up to implementations.

If we're gonna talk "should be up to implementations", then currencies should not have been part of Treasury. Bank permissions and bank accounts should not have been part of Treasury. Heck, even transaction history should not have been part of Treasury.

Same as twitch saying "requiring an oauth key for changing the topic or title of a streamer's stream should be up to the implementation of our API in the language you're gonna develop a bot/oauth app". You know what would be the outcome if they _actually_ do something like this, don't you?

creatorfromhell:
> > it shouldn't be dictated how it's achieved, or the system that should be used.
> 
> Looks like you don't really understand how github reviewing works. I'm suggesting a way to implement it which for me looks more viable and better with the current implementation. If you have a better method please show it. I don't like the thing that I need to implement 1000s of interfaces in order to make an economy provider with 1 currency work.

That's fairly abrupt for a modest conversion I _thought_ was happening. Apparently not. If we're going onto this route, I'm presuming you have no experience in the economy plugin realm, where I do have experience.

> 
> > This would be essentially like registering a permission node to a player without doing so in a permissions plugin. It's an extreme overreach.
> 
> I don't understand your point here.

It's the same premise, the current manager registered and handles currencies without the economy implementation even knowing. Would the same be done in permissions? I would hope not.


> 
> > Yes but it should be up to implementations.
> 
> If we're gonna talk "should be up to implementations", then currencies should not have been part of Treasury. Bank permissions and bank accounts should not have been part of Treasury. Heck, even transaction history should not have been part of Treasury.

This is very reductio ad absurdum. Currencies and transactions are presumed to be in an economy plugin(same as multiworld). I think you're forgetting that majority of economy plugins are well established at this point and if you're implementing things that ought not be implemented by an API do you know what the outcome is? Vault works good enough.

> 
> Same as twitch saying "requiring an oauth key for changing the topic or title of a streamer's stream should be up to the implementation of our API in the language you're gonna develop a bot/oauth app". You know what would be the outcome if they _actually_ do something like this, don't you?

Comparisons to apples and oranges are fun.

MrIvanPlays:
> I'm presuming you have no experience in the economy plugin realm

I've done stuff in the past with economy plugins and I don't wanna go with "implement this and that and that too oh and that too otherwise it won't compile and that other thing and that small thing and that other other thing and yes this also requires implementing that".

The point of Treasury is to offer as much as possible without bloating the API as simple as possible and simple not just for plugins using the API, but for providers too.

> the ~~current~~currency manager registered and handles currencies without the economy implementation even knowing. 

Why should it know? In my opinion it shouldn't care what a plugin does with a currency until they actually use it for a transaction. 

> This is very reductio ad absurdum. 

You wanted "should be up to implementations" things and I listed them.

creatorfromhell:
> > I'm presuming you have no experience in the economy plugin realm
> 
> I've done stuff in the past with economy plugins and I don't wanna go with "implement this and that and that too oh and that too otherwise it won't compile and that other thing and that small thing and that other other thing and yes this also requires implementing that".
> 
> The point of Treasury is to offer as much as possible without bloating the API as simple as possible and simple not just for plugins using the API, but for providers too.
This is the point, the whole current currency system is bloatware. It's an entire implementation of a currency manager that shouldn't be inside an API, aka bloat.

> 
> > the ~current~currency manager registered and handles currencies without the economy implementation even knowing.

I still meant current manager here. Thanks for the grammar fix though.
> 
> Why should it know? In my opinion it shouldn't care what a plugin does with a currency until they actually use it for a transaction.
> 
> > This is very reductio ad absurdum.
> 
> You wanted "should be up to implementations" things and I listed them.

MrIvanPlays:
> This is the point, the whole current currency system is bloatware. It's an entire implementation of a currency manager that shouldn't be inside an API, aka bloat.

Why shouldn't it be? 

The `PluginManager` in BungeeCord is a final class, implemented in the API, and in this thinking, it should not be implemented in the API but in BungeeCord's internals. Why complicate stuff?

> This argument is getting nowhere fast, I think the best method is to wait on Jikoo or lokka30 as you'd rather just be abrupt and correct the grammar of people responding than have a decent conversation.

Just wanted to let you know you sound very unprofessional and childish by saying this.

creatorfromhell:
> > This is the point, the whole current currency system is bloatware. It's an entire implementation of a currency manager that shouldn't be inside an API, aka bloat.
> 
> Why shouldn't it be?
> 
> The `PluginManager` in BungeeCord is a final class, implemented in the API, and in this thinking, it should not be implemented in the API but in BungeeCord's internals. Why complicate stuff?

Because it's bloatware.

> 
> > This argument is getting nowhere fast, I think the best method is to wait on Jikoo or lokka30 as you'd rather just be abrupt and correct the grammar of people responding than have a decent conversation.
> 
> Just wanted to let you know you sound very unprofessional and childish by saying this.
Not really, I'm dealing with someone who goes on unprofessional drunk rants in other topics then decides it's a good idea to correct grammar in a quoted response. That behavior has no business being in a civilized conversion. No to mention this response included. If this is the way you conduct yourself in everyday life I'm sorry.

MrNemo64:
Here I'm on @creatorfromhell's side. An API is ideally a collection of interfaces that an aplication gives to the consumer to implement.
I think we did a good job with the CurrencyManager but having the inplementation in the API defeats the purpouse of the API

> Bukkit is also forcing the plugin manager onto plugins. Bukkit is also forcing you to create plugin.yml so your plugin runs. Bukkit is also forcing you to implement `CommandExecutor` in order to create a command and it is also forcing you to define it in the plugin.yml. Bukkit is forcing you to implement `Listener` so you can register event handlers.

Bukkit is telling you how to interact with their server. They specify a set of rules (the interfaces) that guarantee that your code will run on every Bukkit server.

We, as the designers of an API that we hope everyone will use, are only entitled to define the set of rules that providers and consumers have to follow and hopelully fulfill everyones needs. We don't tell them _how_ they have to follow them, just what they have to follow.

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
```diff
@@ -1,236 +1,141 @@
...
-     * @author lokka30, MrIvanPlays
+     * @return The currency's default number of decimal digits when formatting this currency.
+     * @author creatorfromhell
      * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#v1_0 v1.0}
      */
-    @NotNull
-    public String getPrimaryCurrencyName() {
-        return names[0];
-    }
+    int precision();
```
lokka30, line 62: Should we prefix methods like `precision()` and `displayNamePlural()` with `get`?

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
```diff
@@ -1,236 +1,141 @@
...
+     * @param formatted The formatted string to be converted to double form.
+     * @param subscription The {@link EconomySubscriber} accepting the resulting {@link Double} that
+     *                     represents the deformatted amount of the formatted String.
+     * @author creatorfromhell
      * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#v1_0 v1.0}
      */
-    protected void setConversionCoefficient(double conversionCoefficient) {
-        this.conversionCoefficient = conversionCoefficient;
-    }
+    void deformat(@NotNull String formatted, @NotNull EconomySubscriber<Double> subscription);
```
lokka30, line 109: I believe `parse` could be a better method name. Uncertain on that

@MrIvanPlays what do you think about this change?

core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/economy/EconomyInfoSub.java
```diff
@@ -56,7 +56,7 @@ public void execute(@NotNull final CommandSource sender, @NotNull final String l
                                     .getSupportedOptionalEconomyApiFeatures()
                                     .contains(OptionalEconomyApiFeature.BUKKIT_TRANSACTION_EVENTS))
                     ),
-                    placeholder("primary-currency", provider.getPrimaryCurrency().getPrimaryCurrencyName())
```
lokka30, line 59: I think we should use the currency ID here rather than the currency's display name.

lokka30 (changes requested):

lokka30 (approved):
Awesome - very satisfied with the state of this PR. Thank you very much

creatorfromhell:
> Awesome - very satisfied with the state of this PR. Thank you very much

Cheers. You're welcome.

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
MrIvanPlays: I believe currencies as of right now just need 1-2 small modifications to fulfill OP's needs rather an rewrite.

lokka30:
@Jikoo - whenever you are comfortable, please take a look at this PR :)

core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/economy/migrate/MigrationEconomy.java
```diff
@@ -98,4 +168,49 @@ public void retrieveBankAccountIds(@NotNull EconomySubscriber<Collection<UUID>>
...
+     *                     attempted {@link Currency} registration with an {@link Boolean}.
+     *                     This will be {@link Boolean#TRUE} if it was registered, otherwise
+     *                     it'll be {@link Boolean#FALSE}.
+     *
+     * @author creatorfromhell
+     * @since {@link EconomyAPIVersion#v1_0 v1.0}
+     */
+    @Override
+    public void registerCurrency(@NotNull final Currency currency, @NotNull final EconomySubscriber<Boolean> subscription) {
+subscription.fail(new EconomyException(FailureReason.MIGRATION, "Cannot register currencies during migration!"));
```
Jikoo, line 213: Indentation here.

Jikoo (approved):
Looks good to me. I like currencies not having UUIDs - as much as they may be unique, the unique ID should be easily user-referenced for configuration/commands. Making users determine UUIDs seems like an unnecessary pain.

lokka30:
Thanks Jikoo!

@MrIvanPlays what is your opinion on this PR?

MrIvanPlays:
It has not changed.

lokka30:
I think it would be best if you could list each issue you have with this PR so we can pinpoint them and respond to them.

Doing so, we must have a civil discussion - avoid the personal attacks like last time.

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
```diff
@@ -1,236 +1,141 @@
...
-        this.startingBalance = startingBalance == null ? ($) -> 0d : startingBalance;
-    }
+    /**
+     * Get's the currency's decimal character to be used for formatting purposes.
+     *
+     * @return The currency's decimal character.
+     * @author creatorfromhell
+     * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#v1_0 v1.0}
+     */
+    char getDecimal();
```
MrNemo64, line 35: The decimal character isn't necesarily the same for everyone. It may be `.` or `,` depending on who the formatted string is shown to.

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
```diff
@@ -1,236 +1,141 @@
...
+     * @return The formatted text.
+     * @author creatorfromhell
      * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#v1_0 v1.0}
      */
-    @NotNull
-    public String formatBalance(double amount, @NotNull Locale locale) {
-        Objects.requireNonNull(locale, "locale");
-        return balanceFormatter.apply(amount, locale);
-    }
+    String format(double amount);
```
MrNemo64, line 129: I think this should also accept a `@Nullable Player player` as depending on the player who is going to be shown the amount diferent strings will be generated. For example: `1.000.254,125` or `1,000,254.125` or the provider allows each player to decide the format that the amount is shown and the string ends up being `1 000 254.125` for player A and `1M` for player B

MrNemo64 (changes requested):
We have a method to parse an amount of a given currency but we are missing a method that given a string returns the currency and the amount.

MrIvanPlays:
@lokka30 i already pointed them out.

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
lokka30: > The decimal character isn't necesarily the same for everyone. It may be `.` or `,` depending on who the formatted string is shown to.

That's true - perhaps this method is unnecessary since the format method has a `Locale` argument anyways.

lokka30:
> @lokka30 i already pointed them out.

Please put them in bullet points.

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
lokka30: Ooh, good spotting. Looks like the new method doesn't have the Locale argument. Should we use the Locale or Player? Player being handled by the provider of course.

MrIvanPlays:
> The whole idea of the current implementation is for ease of currency creation not just by the economy provider, but by plugins using the API.
> 
> > Forcing a default formatting system onto economy plugins.
> 
> Nope. Currently the `Currency` class accepts a BiFunction<Double, Locale, String> for formatting a currency.
> 
> > Forcing the conversion standards inside Treasury onto economy plugins.
> 
> We'd be better off changing the conversion coefficient to a BiFunction<Double, Currency, Double> for independent conversion rather than nuking everything and redoing it in the way we wanted to avoid.
> 
> > Forcing the currency manager inside Treasury onto economy plugins
> 
> Bukkit is also forcing the plugin manager onto plugins. Bukkit is also forcing you to create plugin.yml so your plugin runs. Bukkit is also forcing you to implement `CommandExecutor` in order to create a command and it is also forcing you to define it in the plugin.yml. Bukkit is forcing you to implement `Listener` so you can register event handlers.
> 
> > Forcing currency UUIDs that are used nowhere inside the API onto economy plugins.
> 
> Just because they are not used in the treasury plugin does not mean they won't be used _at all_. Currency UUIDs are a way to identify a currency, and a way to store more data with less memory in a set, list, etc.

@lokka30

lokka30:
Not a quote from a conversation :) I've already read that. What exactly is wrong with this PR?

creatorfromhell:
> We have a method to parse an amount of a given currency but we are missing a method that given a string returns the currency and the amount.

Won't let me respond to the conversation, but I see where you're coming from about the decimal character. The only thing is, if we base it off of real-life situations, decimals are tied to currencies not individuals.

MrNemo64:
> > We have a method to parse an amount of a given currency but we are missing a method that given a string returns the currency and the amount.
> 
> Won't let me respond to the conversation, but I see where you're coming from about the decimal character. The only thing is, if we base it off of real-life situations, decimals are tied to currencies not individuals.

I'm not saying to base it off of real-life situations but to let the provider decide what character is used for each player. May be the same for everyone may not be. The `format` method would be the one saying it.

lokka30:
> > We have a method to parse an amount of a given currency but we are missing a method that given a string returns the currency and the amount.
> 
> Won't let me respond to the conversation, but I see where you're coming from about the decimal character. The only thing is, if we base it off of real-life situations, decimals are tied to currencies not individuals.

A server which houses users of different languages would probably like to make each currency be able to adapt its formatting according to an individual player's language :)

creatorfromhell:
> > > We have a method to parse an amount of a given currency but we are missing a method that given a string returns the currency and the amount.
> > 
> > 
> > Won't let me respond to the conversation, but I see where you're coming from about the decimal character. The only thing is, if we base it off of real-life situations, decimals are tied to currencies not individuals.
> 
> I'm not saying to base it off of real-life situations but to let the provider decide what character is used for each player. May be the same for everyone may not be. The `format` method would be the one saying

> > > We have a method to parse an amount of a given currency but we are missing a method that given a string returns the currency and the amount.
> > 
> > 
> > Won't let me respond to the conversation, but I see where you're coming from about the decimal character. The only thing is, if we base it off of real-life situations, decimals are tied to currencies not individuals.
> 
> A server which houses users of different languages would probably like to make each currency be able to adapt its formatting according to an individual player's language :)

Sure, which I agree should be added, but what about default cases? I'm sure they'd like to have an option for a non "." default?

lokka30:
> > A server which houses users of different languages would probably like to make each currency be able to adapt its formatting according to an individual player's language :)
> 
> Sure, which I agree should be added, but what about default cases? I'm sure they'd like to have an option for a non "." default?

Oh, I don't prefer either system over the other, I would rather have both.

lokka30 (approved):

MrIvanPlays (approved):

Issue #76, opened 2021-12-18
Account System Overhaul


creatorfromhell:
The account system as it stands is kind of too similar to Vault's, what I mean by this is why are there just player accounts and bank accounts? While I understand it's just a naming scheme, it gives off a false sense that bank accounts are solely that, accounts for banking systems.

This is why I believe there should be a different Account Structure. It should look something like:

Account
- PlayerAccount (for players)
   - utilizes a UUID
- GenericAccount( for non-players)
  - Utilizes a String identifier
  
  
     While I understand that the likelihood is near impossible of getting a duplicate UUID if we use this system for plugins such as Towny and Factions, it still provides another headache, which is on the server operator's end, which is if they wish to modify one of the balances for a town or faction directly in the database, they won't be able to find it based on a UUID easily.
     
     Also, we are again forcing something onto economy providers that should be up to them to handle appropriately. If they want to store a UUID for a town they should be able to, but if they don't we shouldn't force it onto them.
     
     The last point to this is the fact that under Minecraft convention, we are essentially calling non-player accounts(bank accounts), a Living Entity by providing a UUID to them. 
     
     
     Another point, the account access system should be attached to the Account interface itself. This would allow for the flexibility of player accounts being able to be shared, but not forcing it. It also doesn't restrict it to solely generic accounts.

PR #78, opened 2021-12-18
Account System overhaul. Addresses #76.

creatorfromhell:
This PR addresses a couple of issues.

First, it addresses the issue that BankAccount is misidentified as accounts that are only suitable for bank-related functions when they should be utilized for any account that is not a PlayerAccount. That's why they get renamed to GenericAccounts in order to create a clear inference that they are anything that is not a player account.

Secondly, this addresses the fact that not all accounts should be associated with a UUID. While it is true that the odds of two UUIDs being generated are the same, the fact remains that most economy solutions treat non-player accounts without a UUID. For instance, towny, kingdoms, and lands accounts are prefixed-Name. For instance, a town account is town-TownName in towny. This is something that should be kept in mind, as it has also been utilized as the standard for other platforms such as Sponge. If this truly is supposed to be a platform-independent solution then thought needs to be put into making it as flexible as possible, which in Minecraft terms means players have a UUID and non-players do not.

Lastly, and most insignificant, I changed CONSULT permissions to BALANCE, because why overcomplicate it? It should be straightforward.


lokka30 (commented):
I am very unsure about this change.

I am tilting on agreeing with:
* Making `GenericAccount`s based on String identifiers instead of UUIDs. Previously all the economy providers needed to do was have a database mapping the two but I guess it's okay to switch to Strings.

I disagree with;
* The name `GenericAccount`: it could wrongly suggest that the class is a generic class. Previously, Treasury named `NonPlayerAccount` - perhaps this is better?
* The removal of bank-specific methods in the EconomyProvider file. Player and bank/non-player accounts are created differently
* 'Bank Account permissions' now being 'Account permissions' - they're meant to be for just banks


Spotted issues:
* `BankAccountTransactionEvent` should be `GenericAccountTransactionEvent` in this case

I agree with the renaming of `CONSULT` though :+1:

Thanks for the PR :)

creatorfromhell:
> I am very unsure about this change.
> 
> I am tilting on agreeing with:
> 
> * Making `GenericAccount`s based on String identifiers instead of UUIDs. Previously all the economy providers needed to do was have a database mapping the two but I guess it's okay to switch to Strings.
> 
> I disagree with;
> 
> * The name `GenericAccount`: it could wrongly suggest that the class is a generic class. Previously, Treasury named `NonPlayerAccount` - perhaps this is better?
> * The removal of bank-specific methods in the EconomyProvider file. Player and bank/non-player accounts are created differently
> * 'Bank Account permissions' now being 'Account permissions' - they're meant to be for just banks
> 
> Spotted issues:
> 
> * `BankAccountTransactionEvent` should be `GenericAccountTransactionEvent` in this case
> 
> I agree with the renaming of `CONSULT` though 👍
> 
> Thanks for the PR :)

Here's the reasoning behind this. Having dedicated methods for bank accounts doesn't make sense, especially when all non player accounts are possible to have methods. I.e. town accounts have multiple methods, faction accounts, and this is why it shouldn't be restricted to a specific function.

lokka30:
> > I am very unsure about this change.
> > I am tilting on agreeing with:
> > 
> > * Making `GenericAccount`s based on String identifiers instead of UUIDs. Previously all the economy providers needed to do was have a database mapping the two but I guess it's okay to switch to Strings.
> > 
> > I disagree with;
> > 
> > * The name `GenericAccount`: it could wrongly suggest that the class is a generic class. Previously, Treasury named `NonPlayerAccount` - perhaps this is better?
> > * The removal of bank-specific methods in the EconomyProvider file. Player and bank/non-player accounts are created differently
> > * 'Bank Account permissions' now being 'Account permissions' - they're meant to be for just banks
> > 
> > Spotted issues:
> > 
> > * `BankAccountTransactionEvent` should be `GenericAccountTransactionEvent` in this case
> > 
> > I agree with the renaming of `CONSULT` though 👍
> > Thanks for the PR :)
> 
> Here's the reasoning behind this. Having dedicated methods for bank accounts doesn't make sense, especially when all non player accounts are possible to have methods. I.e. town accounts have multiple methods, faction accounts, and this is why it shouldn't be restricted to a specific function.
> 
> Also when it comes to the removal of generic specific methods it only makes sense if an account is no different than a non player account it terms of identification there shouldn't be similar methods for both. This provides the ability for the economy provider to have new account types without worrying about how a consumer will work with them.

I'm still very uncertain about this. And, what did you think about my other comments? :)

I'll give it some time in case anyone wants to comment before I try make a deeper analysis on this.

creatorfromhell:
> > > I am very unsure about this change.
> > > I am tilting on agreeing with:
> > > 
> > > * Making `GenericAccount`s based on String identifiers instead of UUIDs. Previously all the economy providers needed to do was have a database mapping the two but I guess it's okay to switch to Strings.
> > > 
> > > I disagree with;
> > > 
> > > * The name `GenericAccount`: it could wrongly suggest that the class is a generic class. Previously, Treasury named `NonPlayerAccount` - perhaps this is better?
> > > * The removal of bank-specific methods in the EconomyProvider file. Player and bank/non-player accounts are created differently
> > > * 'Bank Account permissions' now being 'Account permissions' - they're meant to be for just banks
> > > 
> > > Spotted issues:
> > > 
> > > * `BankAccountTransactionEvent` should be `GenericAccountTransactionEvent` in this case
> > > 
> > > I agree with the renaming of `CONSULT` though 👍
> > > Thanks for the PR :)
> > 
> > 
> > Here's the reasoning behind this. Having dedicated methods for bank accounts doesn't make sense, especially when all non player accounts are possible to have methods. I.e. town accounts have multiple methods, faction accounts, and this is why it shouldn't be restricted to a specific function.
> > Also when it comes to the removal of generic specific methods it only makes sense if an account is no different than a non player account it terms of identification there shouldn't be similar methods for both. This provides the ability for the economy provider to have new account types without worrying about how a consumer will work with them.
> 
> I'm still very uncertain about this. And, what did you think about my other comments? :)
> 
> I'll give it some time in case anyone wants to comment before I try make a deeper analysis on this.

The other comments are fair and I'll change accordingly. I still stand by the fact that account permissions should be allowed on all non player accounts, because this is the most flexible option that would allow for the most future proof system, in my opinion. Also, to further explain why it's better to return Account vs GenericAccount. If an economy provider wishes they could make new account classes such as TownAccount and return it without needing new methods. This allows for lots of flexibility when it comes to API design and limits restrictions.

MrIvanPlays (approved):
what lokka mentioned I agree with, otherwise LGTM

MrNemo64:
> The other comments are fair and I'll change accordingly. I still stand by the fact that account permissions should be allowed on all non player accounts, because this is the most flexible option that would allow for the most future proof system, in my opinion. Also, to further explain why it's better to return Account vs GenericAccount. If an economy provider wishes they could make new account classes such as TownAccount and return it without needing new methods. This allows for lots of flexibility when it comes to API design and limits restrictions.

If we move the permissions methods to Account then GenericAccount is the same as Account meaning that we can remove it and we don't have the problem of the name

lokka30:
> > The other comments are fair and I'll change accordingly. I still stand by the fact that account permissions should be allowed on all non player accounts, because this is the most flexible option that would allow for the most future proof system, in my opinion. Also, to further explain why it's better to return Account vs GenericAccount. If an economy provider wishes they could make new account classes such as TownAccount and return it without needing new methods. This allows for lots of flexibility when it comes to API design and limits restrictions.
> 
> If we move the permissions methods to Account then GenericAccount is the same as Account meaning that we can remove it and we don't have the problem of the name

I believe we should keep NonPlayerAccount even if it implements Account with no changes. It's still a specific type of account, but it doesn't add any features upon it.

lokka30 (approved):

lokka30:
@Jikoo would you be able to check this PR? If not, no worries.

@creatorfromhell are we ready to merge this once Jikoo replies? Thank you very much for the PR!

creatorfromhell:
> @Jikoo would you be able to check this PR? If not, no worries.
> 
> @creatorfromhell are we ready to merge this once Jikoo replies? Thank you very much for the PR!

Sure the only question left is Are the permissions methods staying in NonPlayerAccount or moving to Account to be able to be utilized in player accounts too?

lokka30:
> > @Jikoo would you be able to check this PR? If not, no worries.
> > @creatorfromhell are we ready to merge this once Jikoo replies? Thank you very much for the PR!
> 
> Sure the only question left is Are the permissions methods staying in NonPlayerAccount or moving to Account to be able to be utilized in player accounts too?

I believe we can move the permissions to use on all accounts. Although, one thing that should be firm is that Player accounts have all permissions and one member only (being the player itself).

lokka30 (approved):

Jikoo (approved):
Beyond naming nitpick, looks good to me.

I don't particularly like the shift away from UUIDs. I understand that most systems tend to use UUIDs exclusively for players, but that doesn't mean they're correct to do so. Who cares if a player also has a unique identifier? All accounts should have a unique identifier, it just so happens that player accounts' identifiers should also match their owners' identifiers. Why make the identifier for other accounts a String when UUIDs exist? I guess the counterpoint is that the API user is still free to use UUIDs in string form like how Towny surely will, but it seems weird to use a separate construct for a unique ID when that is what a UUID is for. That was why we added the optional name, user friendliness.

lokka30:
> Beyond naming nitpick, looks good to me.
> 
> I don't particularly like the shift away from UUIDs. I understand that most systems tend to use UUIDs exclusively for players, but that doesn't mean they're correct to do so. Who cares if a player also has a unique identifier? All accounts should have a unique identifier, it just so happens that player accounts' identifiers should also match their owners' identifiers. Why make the identifier for other accounts a String when UUIDs exist? I guess the counterpoint is that the API user is still free to use UUIDs in string form like how Towny surely will, but it seems weird to use a separate construct for a unique ID when that is what a UUID is for. That was why we added the optional name, user friendliness.

I have the exact same opinion. Although, I don't mind.

MrNemo64:
> I believe we can move the permissions to use on all accounts. Although, one thing that should be firm is that Player accounts have all permissions and one member only (being the player itself).

At this point this is left for the provider to define. The player associated with the PlayerAccount will always have all permissions but depending on what the provider wants, other players may have permissions too and I don't think there is anything wrong with that.

api/src/main/java/me/lokka30/treasury/api/economy/account/NonPlayerAccount.java
```diff
@@ -0,0 +1,22 @@
...
+
+import java.util.Collection;
+import java.util.Map;
+import java.util.UUID;
+import me.lokka30.treasury.api.economy.response.EconomySubscriber;
+import me.lokka30.treasury.api.misc.TriState;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * A Generic account is any non-player account.
```
MrNemo64, line 15: No longer called like that.
Since the NonPlayerAccount exists only to diferenciate player and non player accounts I think that should be mentioned on the description of the class

api/src/main/java/me/lokka30/treasury/api/economy/account/SharedAccountPermission.java
```diff
@@ -11,42 +11,42 @@
...
- * Enum that holds the permissions of a {@link BankAccount}.
+ * Enum that holds the permissions of an {@link Account} that is shared among multiple players.
  *
  * @author MrNemo64
- * @see BankAccount
+ * @see Account
  * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#v1_0 v1.0}
  */
-public enum BankAccountPermission {
+public enum SharedAccountPermission {
```
MrNemo64, line 20: By calling this `SharedAccountPermission` and being used by all bank accounts we are saying that PlayerAccounts are also shared accounts, whitch may or may not be true, it is left for the provider to specify.

api/src/main/java/me/lokka30/treasury/api/economy/response/FailureReason.java
```diff
@@ -68,31 +72,31 @@
...
     ALREADY_MEMBER_OF_BANK_ACCOUNT,
 
     /**
      * A constant representing failure due to a user already being an owner of a
-     * {@link me.lokka30.treasury.api.economy.account.BankAccount BankAccount}.
+     * {@link NonPlayerAccount BankAccount}.
      *
      * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#v1_0 v1.0}
      */
     ALREADY_OWNER_OF_BANK_ACCOUNT,
```
MrNemo64, line 87: Owners no longer exist in the context of bank accounts. We miss the removal of this failure entry

api/src/main/java/me/lokka30/treasury/api/economy/response/FailureReason.java
```diff
@@ -68,31 +72,31 @@
 
     /**
      * A constant representing failure due to a user already being a member of a
-     * {@link me.lokka30.treasury.api.economy.account.BankAccount BankAccount}.
+     * {@link NonPlayerAccount BankAccount}.
      *
      * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#v1_0 v1.0}
      */
     ALREADY_MEMBER_OF_BANK_ACCOUNT,
```
MrNemo64, line 79: No longer called `BankAccount`

api/src/main/java/me/lokka30/treasury/api/economy/response/FailureReason.java
```diff
@@ -68,31 +72,31 @@
 
     /**
      * A constant representing failure due to a user already being a member of a
-     * {@link me.lokka30.treasury.api.economy.account.BankAccount BankAccount}.
+     * {@link NonPlayerAccount BankAccount}.
```
MrNemo64, line 75: No longer called `BankAccount`

api/src/main/java/me/lokka30/treasury/api/economy/response/FailureReason.java
```diff
@@ -68,31 +72,31 @@
...
      *
      * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#v1_0 v1.0}
      */
     ALREADY_MEMBER_OF_BANK_ACCOUNT,
 
     /**
      * A constant representing failure due to a user already being an owner of a
-     * {@link me.lokka30.treasury.api.economy.account.BankAccount BankAccount}.
+     * {@link NonPlayerAccount BankAccount}.
```
MrNemo64, line 83: No longer called `BankAccount`

api/src/main/java/me/lokka30/treasury/api/economy/response/FailureReason.java
```diff
@@ -68,31 +72,31 @@
...
      *
      * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#v1_0 v1.0}
      */
     ALREADY_OWNER_OF_BANK_ACCOUNT,
 
     /**
      * A constant representing failure due to a user already not being a member of a
-     * {@link me.lokka30.treasury.api.economy.account.BankAccount BankAccount}.
+     * {@link NonPlayerAccount BankAccount}.
```
MrNemo64, line 91: No longer called `BankAccount`

api/src/main/java/me/lokka30/treasury/api/economy/response/FailureReason.java
```diff
@@ -68,31 +72,31 @@
...
      *
      * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#v1_0 v1.0}
      */
     ALREADY_NOT_MEMBER_OF_BANK_ACCOUNT,
```
MrNemo64, line 95: No longer called `BankAccount`

api/src/main/java/me/lokka30/treasury/api/economy/response/FailureReason.java
```diff
@@ -68,31 +72,31 @@
...
 
     /**
      * A constant representing failure due to a user already not being an owner of a
-     * {@link me.lokka30.treasury.api.economy.account.BankAccount BankAccount}.
+     * {@link NonPlayerAccount BankAccount}.
```
MrNemo64, line 99: No longer called `BankAccount`

core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/economy/migrate/GenericAccountMigrator.java
```diff
@@ -93,7 +94,7 @@ public void fail(@NotNull final EconomyException exception) {
 
     @Override
     public @NotNull AtomicInteger getSuccessfulMigrations(@NotNull MigrationData migration) {
-        return migration.bankAccountsProcessed();
+        return migration.genericAccountsProcessed();
```
MrNemo64, line 97: No longer called `GenericAccount`

core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/economy/migrate/GenericAccountMigrator.java
```diff
@@ -12,79 +12,80 @@
...
     @Override
     public @NotNull String getBulkFailLog(@NotNull Throwable throwable) {
         return "Unable to fetch bank account UUIDs for migration: " + throwable.getMessage();
     }
 
     @Override
-    public @NotNull String getInitLog(@NotNull UUID uuid) {
-        return "Migrating bank account of UUID '&b" + uuid + "&7'.";
+    public @NotNull String getInitLog(@NotNull String identifier) {
+        return "Migrating generic account of ID '&b" + identifier + "&7'.";
```
MrNemo64, line 33: No longer called `GenericAccount`

core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/economy/migrate/GenericAccountMigrator.java
```diff
@@ -12,79 +12,80 @@
...
     }
 
     @Override
-    public @NotNull String getErrorLog(@NotNull UUID uuid, @NotNull Throwable throwable) {
-        return "Error migrating bank account UUID '&b" + uuid + "&7': &b" + throwable.getMessage();
+    public @NotNull String getErrorLog(@NotNull String identifier, @NotNull Throwable throwable) {
+        return "Error migrating generic account ID '&b" + identifier + "&7': &b" + throwable.getMessage();
```
MrNemo64, line 38: No longer called `GenericAccount`

core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/economy/migrate/GenericAccountMigrator.java
```diff
@@ -12,79 +12,80 @@
...
     }
 
     @Override
-    public @NotNull BiConsumer<@NotNull EconomyProvider, @NotNull EconomySubscriber<Collection<UUID>>> requestAccountIds() {
-        return EconomyProvider::retrieveBankAccountIds;
+    public @NotNull BiConsumer<@NotNull EconomyProvider, @NotNull EconomySubscriber<Collection<String>>> requestAccountIds() {
+        return EconomyProvider::retrieveGenericAccountIds;
```
MrNemo64, line 43: No longer called `GenericAccount`

MrNemo64 (changes requested):
There are some places that Bank or Generic accounts are referenced. I tryed to mark them all but I may have mist some and there are parts of the code that I couldn't comment (the ones that didn't change)

EDIT: Also you didn't add yourself to the `@author` tag at the methods/classes

api/src/main/java/me/lokka30/treasury/api/economy/account/SharedAccountPermission.java
creatorfromhell: I think this naming scheme better suits it since the methods are in Account now.

MrIvanPlays, line 20: Maybe we can just name it AccountPermission? I'm fine with that name too but what about AccountPermission?

MrIvanPlays:
also you may want to merge `master` into this branch in order to resolve conflicts as we currently cannot merge this. If you can't right now I will.

MrIvanPlays:
Thanks!

Issue #73, opened 2021-12-16
Rework failure reasons


lokka30:
We need to add various failure reasons to the enum.
Feel free to suggest any below and I can add them in once #71 is merged.

Current approved list of additional constants:
```java
DATABASE_ERROR
```

MrNemo64:
Instead of `SQL_ERROR` it should be `DATABASE_ERROR` as a provider may decide to use a no-sql database

MrIvanPlays:
instead of having a enum with fail reasons, convert the enum to a final class or an interface (like the EconomyTransactionInitiator) which holds a message and some other type of data.

PR #79, opened 2021-12-18
Add a more flexible FailureReason system. Addresses #73.

creatorfromhell:
The updated FailureReason system uses an interface and then it's implemented into an enum with the default reasons that we wish to provide internally in Treasury.

### DIFF
--- a/api/src/main/java/me/lokka30/treasury/api/economy/response/TreasuryEconomyFailureReason.java
+++ b/api/src/main/java/me/lokka30/treasury/api/economy/response/TreasuryEconomyFailureReason.java
@@ -1,13 +1,12 @@
-package me.lokka30.treasury.api.economy.response.fail;
+package me.lokka30.treasury.api.economy.response;
 
 import me.lokka30.treasury.api.economy.currency.Currency;
 import me.lokka30.treasury.api.economy.misc.EconomyAPIVersion;
-import me.lokka30.treasury.api.economy.response.EconomySubscriber;
-import me.lokka30.treasury.api.economy.response.FailureReason;
+import me.lokka30.treasury.api.common.response.FailureReason;
 import me.lokka30.treasury.api.economy.transaction.EconomyTransactionInitiator;
 import org.jetbrains.annotations.NotNull;
 
-public enum TreasuryFailureReason implements FailureReason {
+public enum TreasuryEconomyFailureReason implements FailureReason {
 
     /**
      * Use this constant if the method can't be run in any capacity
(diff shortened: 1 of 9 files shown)


api/src/main/java/me/lokka30/treasury/api/economy/response/fail/TreasuryFailureReason.java
```diff
@@ -0,0 +1,343 @@
+package me.lokka30.treasury.api.economy.response.fail;
+
+import me.lokka30.treasury.api.economy.currency.Currency;
+import me.lokka30.treasury.api.economy.misc.EconomyAPIVersion;
+import me.lokka30.treasury.api.economy.response.EconomySubscriber;
+import me.lokka30.treasury.api.economy.response.FailureReason;
+import me.lokka30.treasury.api.economy.transaction.EconomyTransactionInitiator;
+import org.jetbrains.annotations.NotNull;
+
+public enum TreasuryFailureReason implements FailureReason {
```
MrNemo64, line 10: Since this failure reason enum has economy specific entryes and we'll (maybe) have other APIs with their own failure reason this should be named `TreasuryEconomyFailureReason` or `EconomyFailureReason`

MrNemo64 (commented):
Since the doc is delcarated on the FailureReason interface we shouldn't need to copy paste the javadoc

api/src/main/java/me/lokka30/treasury/api/economy/response/fail/TreasuryFailureReason.java
lokka30: To add: the FailureReason should also be moved out of the economy package.

lokka30 (changes requested):
Thank you very much! There is only one thing I could find that we should change , apart from @MrNemo64's comment above.

api/src/main/java/me/lokka30/treasury/api/economy/response/TreasuryEconomyFailureReason.java
```diff
@@ -0,0 +1,342 @@
+package me.lokka30.treasury.api.economy.response;
+
+import me.lokka30.treasury.api.economy.currency.Currency;
+import me.lokka30.treasury.api.economy.misc.EconomyAPIVersion;
+import me.lokka30.treasury.api.common.response.FailureReason;
+import me.lokka30.treasury.api.economy.transaction.EconomyTransactionInitiator;
+import org.jetbrains.annotations.NotNull;
+
+public enum TreasuryEconomyFailureReason implements FailureReason {
```
lokka30, line 9: I believe `Treasury` should be removed from this enum's name

lokka30 (approved):
Thank you very much!
I'll ask the other maintainers what they think.

MrIvanPlays (approved):
as lokka mentioned already, remove `Treasury` from `TreasuryEconomyFailureReason`

Jikoo (approved):
Looks good to me 👍 I agree that removing the Treasury prefix makes sense

MrIvanPlays:
Thanks!

Issue #81, opened 2021-12-21
automation for releasing/verifying pull requests/updating dependencies/pr and issue templates.


portlek:
- [x] Releasing wait for #57 
- [x] Verifying pull requests #84 
- [x] Updating dependencies #83 
  - [x] Need an issue label called `dependencies`
- [x] Issue templates #87 
- [x] P.R. templates

lokka30:
Hi, thanks for the suggestion :)

Do you know of any specific Actions which can accomplish this?

I am unsure about automating releases. Perhaps semi-automation would be better for that.

portlek:
you look at my github action fleet [here](https://github.com/Infumia/InfumiaLib/tree/master/.github/workflows)

portlek:
it has not auto-release feature but it has auto-fill changelog feature for releases so, if any commit come to the master branch, release-drafter takes it and writes the draft-release's change log like that:
![image](https://user-images.githubusercontent.com/12835420/146928276-cbc78c3a-1210-4cbb-8d1f-eebfabc13931.png)
then you go to the edit page:
![image](https://user-images.githubusercontent.com/12835420/146928305-5b78daf2-5ca1-4e80-b973-f67d31d9a8cd.png)
click the publish release, that's it.
![image](https://user-images.githubusercontent.com/12835420/146928121-750de783-36e9-45db-b425-ba5929abc40a.png)

MrIvanPlays:
we're not gonna use github releases. we're gonna use codemc.

either way it would be a good addition for something which updates dependencies.

MrIvanPlays:
do we really need pull request templates?

MrNemo64:
I think templates should be more for outsiders like server owners requesting features/finding bugs, we don't expected to know how the issues work so better to hide them with a template.
PR however, those are created by people that already know so they don't need guide

lokka30:
The first checkbox is out of the scope of this issue (#57).
The last checkbox - PR templates - we've discussed is unnecessary.

Since all of the other checkboxes are completed, is this issue solved?

lokka30:
Thank you very much 🎉

PR #84, opened 2021-12-21
Create build.yml

portlek:
Part of #81, resolves pr verifying.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,6 +1,8 @@
 name: Build
 
 on:
+  push:
+    branches: [ master ]
   pull_request:
     branches: [ master ]
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,10 +11,23 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v2
+      - name: Checkout
+        uses: actions/checkout@v2
 
-      - uses: actions/setup-java@v1
+      - name: Set up JDK
+        uses: actions/setup-java@v2
         with:
-          java-version: 1.8
+          distribution: adopt
+          java-version: 17
 
-      - run: mvn package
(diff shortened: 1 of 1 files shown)


MrIvanPlays:
please make it do so also on master branch commits. thanks

portlek:
@MrIvanPlays i thought if we want to automate the releasing via github actions or codemc things we should separate snapshot workflow from pr verifying.

MrIvanPlays:
> @MrIvanPlays i thought if we want to automate the releasing via github actions or codemc we should separate snapshoting/releasing workflows from pr verifying.

we don't need to. codemc is completely separate from github builds.

MrIvanPlays (approved):

.github/workflows/build.yml
```diff
@@ -0,0 +1,20 @@
...
+    branches: [ master ]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: actions/setup-java@v1
```
Jikoo, line 16: v2 is current, not v1. Should probably also just use current JDK  (unless we want to set up a matrix for multi-version testing later), it's perfectly capable of targeting older versions.

```yaml
      - name: Set up JDK
        uses: actions/setup-java@v2
        with:
          distribution: 'adopt'
          java-version: '17'
```

Jikoo (changes requested):
Should be using cache action to cache Maven repo, speeds up builds tremendously (and reduces load on others' servers) if the repo is reasonably active.

Sample step:
```yaml
      - name: Cache Maven packages
        uses: actions/cache@v2
        with:
          path: ~/.m2/repository
          key: ${{ runner.os }}-maven-${{ hashFiles('**/pom.xml') }}
          restore-keys: |
            ${{ runner.os }}-maven-
```

Jikoo (approved):
LGTM 👍

PR #86, opened 2021-12-21
use same version.

portlek:
discussed in #85


MrIvanPlays:
api version is different than plugin version. reason why the API version is different. 

I'll explain a bit more in the discussion tab you created.

PR #97, opened 2021-12-22
Convert to BigDecimal. Closes #72.

### LINKED ISSUE #72: Convert to BigDecimals
I am unsure whether we should use `double` or `BigDecimal` in our Economy API. Does anyone have arguments for/against BigDecimal?

Our current use of `double` was questioned [here](https://www.spigotmc.org/threads/new-economy-api-looking-for-feedback-from-developers-server-owners.527805/page-5#post-4322734).

## Double
* Sufficient precision (approx. 16 decimal digits).
* Higher performance.

## BigDecimal
* Doesn't suffer from [this](https://stackoverflow.com/questions/3413448/double-vs-bigdecimal) interesting problem.

creatorfromhell:


MrIvanPlays:
Thanks for the contribution!

Unfortunately because your branch looks kinda fucked up, I decided to cherry pick your commit inside Treasury. Please use branches and keep your fork's master branch clean so such stuff do not happen. We want PRs to be marked as "merged" rather than "closed".

PR #98, opened 2021-12-22
Default implement permission methods for player accounts

MrIvanPlays:
@creatorfromhell

### DIFF
--- a/api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
+++ b/api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
@@ -9,7 +9,6 @@ import java.time.Instant;
 import java.time.temporal.Temporal;
 import java.util.Collection;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.UUID;
 import me.lokka30.treasury.api.economy.EconomyProvider;
@@ -349,14 +348,10 @@ public interface Account {
      *
      * @param player       the {@link UUID} of the potential member
      * @param subscription the {@link EconomySubscriber} accepting whether the user is a member
-     * @author lokka30, MrIvanPlays
+     * @author lokka30
      * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#v1_0 v1.0}
      */
-    default void isMember(@NotNull UUID player, @NotNull EconomySubscriber<Boolean> subscription) {
-        Objects.requireNonNull(player, "player");
-        Objects.requireNonNull(subscription, "subscription");
-        hasPermission(player, subscription, AccountPermission.BALANCE); // permission doesn't matter
-    }
+    void isMember(@NotNull UUID player, @NotNull EconomySubscriber<Boolean> subscription);
 
     /**
      * Modifies the state of the specified {@link AccountPermission} {@code permissions} for the specified {@link UUID}
--- a/api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
+++ b/api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
@@ -9,6 +9,7 @@ import java.time.Instant;
 import java.time.temporal.Temporal;
 import java.util.Collection;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.UUID;
 import me.lokka30.treasury.api.economy.EconomyProvider;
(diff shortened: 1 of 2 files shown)


api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
```diff
@@ -348,10 +349,14 @@ default void retrieveTransactionHistory(
...
      * @param subscription the {@link EconomySubscriber} accepting whether the user is a member
-     * @author lokka30
+     * @author lokka30, MrIvanPlays
      * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#v1_0 v1.0}
      */
-    void isMember(@NotNull UUID player, @NotNull EconomySubscriber<Boolean> subscription);
+    default void isMember(@NotNull UUID player, @NotNull EconomySubscriber<Boolean> subscription) {
+        Objects.requireNonNull(player, "player");
+        Objects.requireNonNull(subscription, "subscription");
+        hasPermission(player, subscription, AccountPermission.BALANCE); // permission doesn't matter
```
MrNemo64, line 358: A player is an account member as long as they have at least one permission. By assuming that any member will have at least `BALANCE` we don't follow that description as someone may have the permission to deposit but not consult. If that makes sense or not is left for the provider to define

api/src/main/java/me/lokka30/treasury/api/economy/account/PlayerAccount.java
```diff
@@ -49,6 +67,69 @@
...
+    @Override
+    default void setPermission(
+            @NotNull UUID player,
+            @NotNull TriState permissionValue,
+            @NotNull EconomySubscriber<Boolean> subscription,
+            @NotNull AccountPermission @NotNull ... permissions
+    ) {
+        // do nothing. economy api users and by users I mean plugins which are not the economy
+        // provider plugin shall not be able to modify anything on a player account except do
+        // transactions.
```
MrNemo64, line 130: > The default interaction is that a consumer plugin may not modify the permissions of a player account, although is left for the provider plugin to override this interaction

What about that way of saying it?
Also `subscription` should succed of fail

MrNemo64 (changes requested):

MrNemo64:
I don't know if specifying default methods for permissions on the player account is a good idea.
I understand the logic behind it but I don't know if we should do it.

api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
MrIvanPlays: this means that `Account` won't have a default implementation for `isMember`.

MrNemo64: It could have by checking each permission individually and when you find one that the player has succeeded with true, if no permission is found succeeded with false

MrIvanPlays: > It could have by checking each permission individually and when you find one that the player has succeeded with true, if no permission is found succeeded with false

This kills the spirit of SQL queries if anyone's going to use mysql.

MrNemo64: You can also use `retrievePermissions` and see if at least one is set to true

MrIvanPlays: > You can also use `retrievePermissions` and see if at least one is set to true

why should we load all the permissions when we can just do a simple index check in mysql? the counterpoint of this would be that currently the permissions are only 4, but let it stay as is.

lokka30 (approved):

creatorfromhell:
> @creatorfromhell

I like it, seems fairly straightforward at what it does and should make lives easier for implementations.

Jikoo (approved):

api/src/main/java/me/lokka30/treasury/api/economy/account/PlayerAccount.java
```diff
@@ -27,6 +33,18 @@
  */
 public interface PlayerAccount extends Account {
 
+    /**
+     * Returns a map fulfilled with all {@link AccountPermission} with {@link TriState} values of
+     * {@link TriState#TRUE}
+     */
+    Map<AccountPermission, TriState> ALL_PERMISSIONS_MAP = Collections.unmodifiableMap(new ConcurrentHashMap<AccountPermission, TriState>() {
```
portlek, line 40: This code can be simplified by:
```java
Collections.unmodifiableMap(Arrays.stream(AccountPermission.values())
  .collect(Collectors.toConcurrentMap(permission -> permission, permission -> TriState.TRUE)))
```

PR #113, opened 2022-01-18
Support hex codes as advertised

MrIvanPlays:
The messages.yml says that the format `&#123456` for hex colors is supported whilst it isn't in bukkit. I have implemented that code only for bukkit due to the fact other platforms (such as velocity and sponge, which use adventure) support that format already.


lokka30 (approved):
Thank you very much! Ready to merge whenever you are. 👍🏻

PR #127, opened 2022-01-26
[WIP] BungeeCord support

MrIvanPlays:
This has awfully lot of stuff to do. Thoughts wanted, especially on the way a EconomyProvider would be obtained because BungeeCord does not have a service provider.

### DIFF
--- a/platform/bungee/api/src/main/java/me/lokka30/treasury/api/economy/bungee/misc/BungeeCordEventCaller.java
+++ b/platform/bungee/api/src/main/java/me/lokka30/treasury/api/economy/bungee/misc/BungeeCordEventCaller.java
@@ -1,0 +1,139 @@
+package me.lokka30.treasury.api.economy.bungee.misc;
+
+import java.util.Objects;
+import me.lokka30.treasury.api.economy.account.NonPlayerAccount;
+import me.lokka30.treasury.api.economy.account.PlayerAccount;
+import me.lokka30.treasury.api.economy.event.AccountTransactionEvent;
+import me.lokka30.treasury.api.economy.event.NonPlayerAccountTransactionEvent;
+import me.lokka30.treasury.api.economy.event.PlayerAccountTransactionEvent;
+import me.lokka30.treasury.api.economy.transaction.EconomyTransaction;
+import net.md_5.bungee.api.Callback;
+import net.md_5.bungee.api.ProxyServer;
+import net.md_5.bungee.api.plugin.Cancellable;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * A class which makes it easier with calling Treasury events.
+ * <p>
+ * Story goes the following: BungeeCord does not have any event hierarchy such as Bukkit has. If
+ * you call an event extending another event, what would happen in Bukkit is that the event which
+ * the called event extends will also be called, whilst in BungeeCord it would not! That's why we
+ * have this class ; to call 2 events at the same time easily and combine their outcomes.
+ *
+ * @author MrIvanPlays
+ */
+public final class BungeeCordEventCaller {
+
+    /**
+     * Calls {@link AccountTransactionEvent} and {@link PlayerAccountTransactionEvent}
+     *
+     * @param account     the player account that transaction is made onto
+     * @param transaction the transaction
+     * @param callback    the callback of the event outcome - whether at least one of the events is
+     *                    cancelled.
+     * @author MrIvanPlays
+     */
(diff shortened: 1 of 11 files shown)


lokka30:
> This has awfully lot of stuff to do. Thoughts wanted, especially on the way a EconomyProvider would be obtained because BungeeCord does not have a service provider.

I hope the other platforms have a service provider.
I think the only solutions are
* make the service registration in Treasury's bungee core (not the API of course)
* make the service registration a separate library for bungee and whichever else requires it

platform/bungee/api/src/main/java/me/lokka30/treasury/api/economy/bungee/misc/BungeeCordEventCaller.java
```diff
@@ -0,0 +1,139 @@
...
+
+                callback.done(isCancelled(event, patEvent), null);
+            };
+
+            ProxyServer
+                    .getInstance()
+                    .getPluginManager()
+                    .callEvent(new NonPlayerAccountTransactionEvent(account,
+                            transaction,
+                            pateCallback
```
lokka30, line 113: shouldn't this be `npate`

platform/bungee/api/src/main/java/me/lokka30/treasury/api/economy/bungee/misc/BungeeCordEventCaller.java
```diff
@@ -0,0 +1,139 @@
...
+
+        ProxyServer.getInstance().getPluginManager().callEvent(new AccountTransactionEvent(account,
+                transaction,
+                ateCallback
+        ));
+    }
+
+    private static boolean isCancelled(Cancellable event, Cancellable event1) {
+        boolean cancelled;
+        if (event.isCancelled() && event1.isCancelled()) {
```
lokka30, line 125: probably better just to make this a big return statement

lokka30: also no need for the `cancelled` variable, directly return in the statements

platform/bungee/plugin/pom.xml
```diff
@@ -0,0 +1,125 @@
...
+            <version>${bungeecord.version}</version>
+            <scope>provided</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.jetbrains</groupId>
+            <artifactId>annotations</artifactId>
+            <version>${jetbrains.annotations.version}</version>
+            <scope>provided</scope>
+        </dependency>
+        <!-- TODO: bStats?
```
lokka30, line 114: how come did we comment it out?

lokka30 (commented):
Thank you very much for working on this!

platform/bungee/plugin/pom.xml
MrIvanPlays: I commented it out with a todo because bungee metrics require a new page on bstats.

MrIvanPlays:
> > This has awfully lot of stuff to do. Thoughts wanted, especially on the way a EconomyProvider would be obtained because BungeeCord does not have a service provider.
> 
> 
> 
> I hope the other platforms have a service provider.

Thing is velocity also doesn't have one. The difference is that velocity is willing to get one whilst md_5 is strongly against adding such in bungee.

lokka30:
> > > This has awfully lot of stuff to do. Thoughts wanted, especially on the way a EconomyProvider would be obtained because BungeeCord does not have a service provider.
> > 
> > 
> > I hope the other platforms have a service provider.
> 
> Thing is velocity also doesn't have one. The difference is that velocity is willing to get one whilst md_5 is strongly against adding such in bungee.

Should we add a service registration methods into the Treasury Bungee Core?

MrIvanPlays:
Thats what I did in the Bungee API but I don't think it is the best implementation. I'd like @Jikoo's opinion on this

MrNemo64:
Since Bungee doesn't have the service provider, with your current implementation, we would need a diferent provider class for every api. 
In case we decide to go with the option of creating our own service provider (which I personally can't think of a better way), I think it should work to provide any API, not just the EconomyProvider

MrIvanPlays:
what im also thinking of is our own event handler. this will eliminate the need for per platform apis which would basically mean that treasury api could be implemented also in non minecraft related programs ( if anyone would use it for such )

MrIvanPlays:
looks like im trying to do stuff too soon. closing this and im gonna open a discussion tab about the problems.

PR #137, opened 2022-01-27
Implement service provider

MrIvanPlays:


api/src/main/java/me/lokka30/treasury/api/common/services/Service.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /**
+     * Returns the held service.
+     *
+     * @return service
+     */
+    public @NotNull T get() {
+        return service;
+    }
```
MrNemo64, line 55: Maybe we could implement `Supplier`

MrIvanPlays: Unnecessary.

MrNemo64: Right now this class pretty much is a supplier. Why not implement it?

MrIvanPlays: Oh wait I read that wrong. Thought you meant to replace this with a supplier. Ok we will implement supplier

MrIvanPlays: done

api/src/main/java/me/lokka30/treasury/api/common/services/Service.java
```diff
@@ -0,0 +1,72 @@
...
+ * @author MrIvanPlays
+ */
+public final class Service<T> implements Comparable<Service<?>>, Supplier<T> {
+
+    private final String registrator;
+    private final ServicePriority priority;
+    private final T service;
+
+    public Service(
+            @NotNull String registrator, @NotNull ServicePriority priority, @NotNull T service
```
lokka30, line 24: `service` should probably be `serviceType`, also `Service get()` should probably be `getServiceType()` if you agreed.

lokka30 (changes requested):
Cheers Ivan!

api/src/main/java/me/lokka30/treasury/api/common/services/Service.java
MrIvanPlays: not really. we're providing a service, not a service type. can't agree for both of these things here.

lokka30: Ok. But I think `get` should be `getService` because when combined with optionals it looks like `get().get()`. Nitpick though, feel free to mark this as resolved.

lokka30 (approved):

MrIvanPlays:
@Jikoo will you be able to review soon ?

Jikoo:
Haven't fully reviewed, but this looks breaking due to the removal of several classes and methods. Deleting the internal version should be fine, but the removal of public methods would be a 2.0 bump, they should remain and be deprecated until then.

MrIvanPlays:
> Haven't fully reviewed, but this looks breaking due to the removal of several classes and methods. Deleting the internal version should be fine, but the removal of public methods would be a 2.0 bump, they should remain and be deprecated until then.

this does not remove anything that is api.

MrIvanPlays:
image

Jikoo:
> this does not remove anything that is api.

Gotcha, good.

Jikoo (changes requested):
Lacks backwards compatibility - if a plugin depends on Treasury 1.0 for its API, it will expect services via the Bukkit ServiceManager. That means that all services registered via Treasury 1.1's methods will be entirely unavailable to those plugins. The only way to maintain compatibility is to ignore the 1.1 API and register via the old technique, which will also yield 1.1 compatibility and Treasury's nag. Were I an economy implementer I would actually unregister Treasury's adapter and replace it with one without a nag, then use the old technique.

MrIvanPlays:
> Lacks backwards compatibility - if a plugin depends on Treasury 1.0 for its API, it will expect services via the Bukkit ServiceManager. That means that all services registered via Treasury 1.1's methods will be entirely unavailable to those plugins. The only way to maintain compatibility is to ignore the 1.1 API and register via the old technique, which will also yield 1.1 compatibility and Treasury's nag. Were I an economy implementer I would actually unregister Treasury's adapter and replace it with one without a nag, then use the old technique.

What shall we do about this then? I'm open to any ideas. The nag is the easiest to remove.

MrNemo64:
Maybe make this it's own module and use the platform specific system to get the providers, but if there is no platform specific system then the API would depend on this module

MrIvanPlays:
> Maybe make this it's own module and use the platform specific system to get the providers, but if there is no platform specific system then the API would depend on this module

this doesn't make sense to me. could you be more clear or fix your english? thanks

MrNemo64:
> > Maybe make this it's own module and use the platform specific system to get the providers, but if there is no platform specific system then the API would depend on this module
> 
> this doesn't make sense to me. could you be more clear or fix your english? thanks

Sure! My idea is this: take the service provider API and put it in its own module, `service-provider` for example. Now, if a platform has its own service providing system, use the system of the platform, otherwise the api of the platform would depend on the `service-provider` module and would use it

MrIvanPlays:
> > > Maybe make this it's own module and use the platform specific system to get the providers, but if there is no platform specific system then the API would depend on this module
> > 
> > 
> > this doesn't make sense to me. could you be more clear or fix your english? thanks
> 
> Sure! My idea is this: take the service provider API and put it in its own module, `service-provider` for example. Now, if a platform has its own service providing system, use the system of the platform, otherwise the api of the platform would depend on the `service-provider` module and would use it

that doesn't make the treasury api platform independent, what we target here.

MrIvanPlays:
#142 should be merged first before continuing on this.

Jikoo:
> > Lacks backwards compatibility - if a plugin depends on Treasury 1.0 for its API, it will expect services via the Bukkit ServiceManager. That means that all services registered via Treasury 1.1's methods will be entirely unavailable to those plugins. The only way to maintain compatibility is to ignore the 1.1 API and register via the old technique, which will also yield 1.1 compatibility and Treasury's nag. Were I an economy implementer I would actually unregister Treasury's adapter and replace it with one without a nag, then use the old technique.
> 
> What shall we do about this then? I'm open to any ideas. The nag is the easiest to remove.

I'm not sure honestly, I just really don't like the idea of making a change to services that will quietly break compatibility on a minor version bump. Maybe we should wait on adding events, then add a registration event? That would allow compatibility from either end.

MrIvanPlays:
> > > Lacks backwards compatibility - if a plugin depends on Treasury 1.0 for its API, it will expect services via the Bukkit ServiceManager. That means that all services registered via Treasury 1.1's methods will be entirely unavailable to those plugins. The only way to maintain compatibility is to ignore the 1.1 API and register via the old technique, which will also yield 1.1 compatibility and Treasury's nag. Were I an economy implementer I would actually unregister Treasury's adapter and replace it with one without a nag, then use the old technique.
> > 
> > 
> > What shall we do about this then? I'm open to any ideas. The nag is the easiest to remove.
> 
> I'm not sure honestly, I just really don't like the idea of making a change to services that will quietly break compatibility on a minor version bump. Maybe we should wait on adding events, then add a registration event? That would allow compatibility from either end.

thats why i converted to draft. realized we don't have proper events yet so im gonna make events work first and then continue on this.

MrIvanPlays:
did some work on my computer, will finish it off tomorrow.

MrIvanPlays:
@Jikoo please do a review whenever you can. thanks

MrIvanPlays:
performed tests, backwards compatibility is working flawlessly :)

Jikoo (approved):
Looks excellent 👍

PR #139, opened 2022-02-02
Improvements

lokka30:


MrNemo64:
What more improvements are scoped for this PR? Or are you wrinting as you develop your provider?

lokka30:
> What more improvements are scoped for this PR? Or are you wrinting as you develop your provider?

I haven't been able to come up with any other improvements than the single one in this PR 😛 so I think I'll just merge this now. I haven't started working on my provider yet, but it was just something I noticed whilst I was helping JustEli add the Treasury integration into his resource.
The code in this commit will need to change when #128 is implemented due to the change in the service provider system.

PR #144, opened 2022-02-10
get rid of stupid deprecation warning

MrIvanPlays:
gets rid of this crap
image

### DIFF
--- a/platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/TreasuryBukkit.java
+++ b/platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/TreasuryBukkit.java
@@ -5,7 +5,10 @@
 package me.lokka30.treasury.plugin.bukkit;
 
 import java.io.File;
+import java.util.Arrays;
+import java.util.List;
 import java.util.Optional;
+import java.util.logging.Filter;
 import me.lokka30.treasury.api.common.event.EventExecutorTrackerShutdown;
 import me.lokka30.treasury.api.common.services.Service;
 import me.lokka30.treasury.api.common.services.ServiceProvider;
@@ -41,6 +44,13 @@ public class TreasuryBukkit extends JavaPlugin {
 
     private BukkitTreasuryPlugin treasuryPlugin;
 
+    private static final List<String> ownDeprecatedEventNames = Arrays.asList(
+            me.lokka30.treasury.api.economy.event.AccountEvent.class.getName(),
+            me.lokka30.treasury.api.economy.event.AccountTransactionEvent.class.getName(),
+            me.lokka30.treasury.api.economy.event.NonPlayerAccountTransactionEvent.class.getName(),
+            me.lokka30.treasury.api.economy.event.PlayerAccountTransactionEvent.class.getName()
+    );
+
     /**
      * Run the start-up procedure for the plugin.
      * This is called by Bukkit's plugin manager.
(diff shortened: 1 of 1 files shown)


Jikoo (approved):
👍

MrNemo64:
Wouldn't this block all deprecated event messages? Can we make the filter more precise?

platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/TreasuryBukkit.java
```diff
@@ -50,6 +51,16 @@
      */
     @Override
     public void onEnable() {
+        // get rid of this warning: https://img.mrivanplays.com/jaqzdjmvpz.png .We register a
+        // listener for our events for backwards compatibility. Couldn't bukkit be smarter and not
+        // print it for events which are created by the plugin which is listening??
```
lokka30, line 56: should I submit an issue to Paper about this?

lokka30 (approved):

yannicklamprecht:
```java
    private static final List<Class<?>> ownDeprecatedClasses = Arrays.asList(
            me.lokka30.treasury.api.economy.event.AccountEvent.class,
            me.lokka30.treasury.api.economy.event.AccountTransactionEvent.class,
            me.lokka30.treasury.api.economy.event.NonPlayerAccountTransactionEvent.class,
            me.lokka30.treasury.api.economy.event.PlayerAccountTransactionEvent.class
    );
  ``



```java
getLogger().setFilter(record -> {
            if (ownDeprecatedClasses.stream().anyMatch(aClass -> record.getMessage().contains(aClass.getName()))
                    && record.getMessage().contains("but the event" +
                    " is Deprecated")) {
                return false;
            }
            return oldFilter == null || oldFilter.isLoggable(record);
        });
```

Should do it. Probably

yannicklamprecht (changes requested):
```java
    private static final List<Class<?>> ownDeprecatedClasses = Arrays.asList(
            me.lokka30.treasury.api.economy.event.AccountEvent.class,
            me.lokka30.treasury.api.economy.event.AccountTransactionEvent.class,
            me.lokka30.treasury.api.economy.event.NonPlayerAccountTransactionEvent.class,
            me.lokka30.treasury.api.economy.event.PlayerAccountTransactionEvent.class
    );
  ```



```java
getLogger().setFilter(record -> {
            if (ownDeprecatedClasses.stream().anyMatch(aClass -> record.getMessage().contains(aClass.getName()))
                    && record.getMessage().contains("but the event" +
                    " is Deprecated")) {
                return false;
            }
            return oldFilter == null || oldFilter.isLoggable(record);
        });
```

Should do it. Probably

MrNemo64:
I do like that filter more

yannicklamprecht:
> I do like that filter more. Since the filter will probably run pretty often maybe we should consider more optimal ways of checking?

Logging is done on a separate thread but looking into converting that stream into a "primitive" loop would be a good idea.
First checking for  `but the event is Deprecated` and then looping through the deprecated classes should be better.
Additionally we can just initialize the list with the name of that class. Will comment it in a few.

platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/TreasuryBukkit.java
yannicklamprecht: I'm a fan of adding a class annotation for that.

From that example a missing annotation should not throw a CNFE.
https://www.baeldung.com/classnotfoundexception-missing-annotation

So something like that could be possible in Paper-API:

```java
@Retention(RetentionPolicy.RUNTIME)
    @Target(ElementType.TYPE)
    public @interface DoNotWarnFor {
        Class<?>[] plugins();
    }
    
    @Deprecated
    @DoNotWarnFor(plugins = TreasuryBukkit.class)
    public class SomeEvent extends Event {
        
        @Override
        public @NotNull HandlerList getHandlers() {
            return null;
        }
    }
```

That could be filtered.

yannicklamprecht: But that is the lazy variant, better would be to compare the class loader of the event class with the class loader of there registering class.

lokka30:
Thanks @yannicklamprecht, that solution is much better.

MrIvanPlays:
I will convert the pr with yannick's solution and merge.

platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/TreasuryBukkit.java
MrIvanPlays: Thats just a comment explaining why is that code. What the fuck have you actually did here?

MrNemo64:
I just thought of something, will this block all deprecation warning messages from all plugins or only from treasury?

MrIvanPlays:
> I just thought of something, will this block all deprecation warning messages from all plugins or only from treasury?

treasury only

lokka30:
> I just thought of something, will this block all deprecation warning messages from all plugins or only from treasury?

it blocks from Treasury's logger :) thru `Plugin#getLogger()`

PR #145, opened 2022-02-11
Deprecation and eventual removal of API Versions

lokka30:
I would also like to discuss a target plugin version for the complete removal of all associated classes and methods. :)

I think removing it in v1.2.0 could leave a damaged (albeit little to none) reputation of the library's stability. Perhaps v1.3.0?

### DIFF
--- a/api/src/main/java/me/lokka30/treasury/api/common/event/Cancellable.java
+++ b/api/src/main/java/me/lokka30/treasury/api/common/event/Cancellable.java
@@ -8,7 +8,7 @@ package me.lokka30.treasury.api.common.event;
  * An interface, which can be implemented by events which should be cancellable.
  *
  * @author MrIvanPlays
- * @since {@link me.lokka30.treasury.api.economy.misc.EconomyAPIVersion#V1_1 v1.1}
+ * @since v1.1.0
  */
 public interface Cancellable {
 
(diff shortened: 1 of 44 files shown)


MrIvanPlays:
we will remove deprecated stuff in 2.0 in order to follow semver.

MrIvanPlays (approved):

MrIvanPlays:
> We could give a 5 minor updates time frame to remove deprecated APIs or with every major remove everything that's deprecated although the last option may not give much time to adapt in certain scenarios

im fairly sure we're releasing 1.1 and we're gonna stop the development for a bit ; start helping people to implement treasury support

lokka30:
> > We could give a 5 minor updates time frame to remove deprecated APIs or with every major remove everything that's deprecated although the last option may not give much time to adapt in certain scenarios
> 
> im fairly sure we're releasing 1.1 and we're gonna stop the development for a bit ; start helping people to implement treasury support

^^^ this.

so to remove deprecated stuff, the following factors are at play:
* how many people are using the deprecated code
* how long it has been since the code was marked as deprecated
* how urgent we are to remove the deprecated code

for the API version removal, these are the factors:
* few plugins are using the deprecated code
* code was only marked as deprecated just now so we should wait either a major version or a month or so.
* not urgent since it does not affect anything by using it

MrNemo64:
Sounds about right

Jikoo (approved):
Agree that per semver the deprecated methods should not be removed until 2.0, but there's no reason to not release a 2.0 even if it's exclusively to remove deprecated methods after a sufficient wait.

Issue #146, opened 2022-02-12
Document the new Service & Event APIs


lokka30:
I'll need to document the new service & event APIs before we release v1.1.0.

lokka30:
On hold: waiting for the resolution of #147.

MrIvanPlays:
Already documented

lokka30:
Awesome. Waiting on the result of #147 to close this.

MrIvanPlays:
Well if #147 turns out to be done then the only thing which will need update is the javadocs links on the wiki page. Nothing else

PR #147, opened 2022-02-12
Moving service/event packages

lokka30:
Currently the brand new and brilliant Service and Event APIs that @MrIvanPlays programmed reside in the `api.common` package.

I disagree - these new APIs should be in the `api` package instead, because:
* The `api.common` package is suitable only for code that is shared between multiple of Treasury's APIs, but **not new APIs entirely**.
* New APIs warrant having their own sub-package in the `api` package, they should not be buried under `api.common`.

To resolve this, I have made the following changes:
* renamed `services` to `service`, to pertain with the `economy` and `event` packages not being in plural form.
* moved `service` and `event` packages from `api.common` to `api`.
* edited `package-info` descriptions.

I would like to hear opinions from @MrIvanPlays and @Jikoo if possible. 😄 

Once this is resolved, I can work on #146, and then, release v1.1.0! 🍺 

***

This PR is all ready for review and merge as-is.


MrIvanPlays:
No. I agree renaming services to service, but moving them outta common. No.

lokka30:
> No. I agree renaming services to service, but moving them outta common. No.

Alright. I'll let Jikoo settle this if he has an opinion on it

lokka30:
Tagging other users who may be interested: @MrNemo64 @yannicklamprecht

MrNemo64:
I'm okay with both options

yannicklamprecht (approved):
This api.common thingy wasn't released?
Otherwise it would be a breaking change.

lokka30:
> This api.common thingy wasn't released? Otherwise it would be a breaking change.

There has always been a `common` package inside the `api` package :) is this what you are talking about?

MrIvanPlays:
We haven't officially said that you need to use the version. Although it is deployed to codemc and repo.mrivanplays.com .

Jikoo:
I second Ivan. I thought the whole point of the common packages were that they defined the things that were common to all APIs. If we were to move one thing out we should remove common from all package paths, which would be breaking. I think it could have been a good change prior to release, but it doesn't make sense now.

lokka30:
> I second Ivan. I thought the whole point of the common packages were that they defined the things that were common to all APIs. If we were to move one thing out we should remove common from all package paths, which would be breaking. I think it could have been a good change prior to release, but it doesn't make sense now.

We are yet to release the new APIs, so this PR makes zero breaking changes. In v1.0 the common package only includes utilities, being the FailureReason and TriState interfaces. This is a stark contrast to now holding entire APIs within it.

Jikoo:
> > I second Ivan. I thought the whole point of the common packages were that they defined the things that were common to all APIs. If we were to move one thing out we should remove common from all package paths, which would be breaking. I think it could have been a good change prior to release, but it doesn't make sense now.
> 
> We are yet to release the new APIs, so this PR makes zero breaking changes. In v1.0 the common package only includes utilities, being the FailureReason and TriState interfaces. This is a stark contrast to now holding entire APIs within it. Do you still second Ivan?

Yes. The point of the common subpackages was that they were things useable by all Treasury APIs, though it is somewhat redundant because Treasury will distribute all its APIs in a single archive. If we were to do this, we should also move `FailureReason` and `TriState` to `...api.x` instead of `...api.common.x`.

lokka30:
Alright. I still disagree because this is nothing like what the common package was meant to handle. Regardless, I'll go with the majority vote here. I hope this does not bite us in the back in the future because this is pretty much irreversible.

@MrIvanPlays all good to release once I rename `services` -> `service`?

MrIvanPlays:
Looks like you @lokka30 completely misunderstood the idea of the common package. Jikoo explained it pretty much.

lokka30:
(I have not really proof-read this so let me know if there are any holes in this)

@Jikoo @MrIvanPlays (and @MrNemo64, in case you're interested too)

I still believe there is a strong misunderstanding here. It could be me, I don't know, but I'll do what I can to try make myself clear here because this will affect the organisation of API packages in the future. I would appreciate if everyone could give this a bit of thought because misorganisation is irreversible, considering the stability of the library. I'll give this one last shot.

***

Ivan, although it seems your opinion is somewhat aligned with Jikoo's, you have not provided any reasoning of your own here for why you believe what you do. So unfortunately I am only able to reply to Jikoo here. An opinion with zero foundation creates more confusion.

Here goes...

> The point of the common subpackages was that they were things usable by all Treasury APIs ...

Correct. :)

> ... though it is somewhat redundant because Treasury will distribute all its APIs in a single archive.  If we were to do this, we should also move FailureReason and TriState to ...api.x instead of ...api.common.x.

This is where the misunderstanding has occurred:

No - the difference between `FailureReason` and `TriState` to the `service` and `event` packages, are that the first two are utility classes that do not present significance to anyone unless they are using methods in any of the APIs that require them.

For example, the Economy package uses TriStates and FailureReasons. Do people have any primary purpose to use them? No, they are only using them because the methods in (or branching out of) the EconomyProvider require them to.

The service and event packages are **significant** APIs. Yes, like code that resides in the common package, they are used by multiple APIs (well, currently only Economy of course). But this does not mean that significant APIs should follow suit because code already exists in it - code which is insignificant. We're putting two entire APIs inside the common package, when it was only meant to house common utilities shared between APIs.

I disagree to the removal of the common package and instead using inter-dependence for these utilities. It would be weird if the Permissions API depended upon something that was nested inside the Economy API. Completely different things, and the code could become spaghetti-like (relatively) if we were to go this route. I think we should avoid inter-API dependence where we can (of course with exception to those utilizing events/services).

***

If you're done with me trying to resurface this, then plainly let me know, I really can't be bothered wasting more time on this especially if it will go nowhere. As I commented just recently in another issue, all I'm trying to do here is make improvements to suit the long-term survival of the current code, I am not dying on any of these hills I'm creating. I will back down and let things be if that's how it is. But I do feel the direction this is currently going will be detrimental. I don't waste my time (and I don't want to waste your time) by rambling on about insignificant things, so I wouldn't persist here if I felt this issue wasn't important.

MrIvanPlays:
The events and services apis are separate, that's correct. But the thing is they are not completely mainstream treasury apis, the reason of why I put them in common. If we follow your way of seeing things, the TriState can also be considered a full api because it can be used completely separately from treasury's apis, hence in your logic it should also have its own package.

lokka30:
> The events and services apis are separate, that's correct. But the thing is they are not completely mainstream treasury apis, the reason of why I put them in common. If we follow your way of seeing things, the TriState can also be considered a full api because it can be used completely separately from treasury's apis, hence in your logic it should also have its own package.

I appreciate your response. But as I said, `TriState` and `FailureReason` are incomparable to the APIs, so my logic does not apply. Those are purely utility classes. What do you think now?

MrIvanPlays:
Can't change my opinion on this. Im gonna let democracy win - if you change jikoo's opinion then do it but I am rock hard on this.

Jikoo:
Frankly the reason I'm against it is because the common package exists. Imo `common` should have been named `util` and the only reason I didn't notice and say something earlier was that I was under the impression that Treasury's future APIs would be separate bundles (hence my confusion on versioning as well). At this point it's too late to rename common, so all common things (even if they're not utilities and might have been better served by separate packages) should end up there. It's a consistency vs best practice thing for me. I would rather see Treasury be consistent than use the most brief and descriptive package names.

lokka30:
Thank you very much for the reply Jikoo :) democracy has decided against this change.

PR #158, opened 2022-02-17
Start working on PAPI hook

### LINKED ISSUE #156: Add PlaceholderAPI placeholders for the Economy API

MrIvanPlays:
Closes #156

### DIFF
--- a/platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/EconomyHook.java
+++ b/platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/EconomyHook.java
@@ -1,0 +1,73 @@
+/*
+ * This file is/was part of Treasury. To read more information about Treasury such as its licensing, see <https://github.com/lokka30/Treasury>.
+ */
+
+package me.lokka30.treasury.plugin.bukkit.hooks;
+
+import java.math.BigDecimal;
+import java.text.DecimalFormat;
+import java.util.Optional;
+import me.lokka30.treasury.api.common.service.Service;
+import me.lokka30.treasury.api.common.service.ServiceRegistry;
+import me.lokka30.treasury.api.economy.EconomyProvider;
+import me.lokka30.treasury.api.economy.account.PlayerAccount;
+import me.lokka30.treasury.api.economy.response.EconomySubscriber;
+import org.bukkit.OfflinePlayer;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class EconomyHook implements TreasuryPAPIHook {
+
+    private EconomyProvider provider;
+    private final DecimalFormat format = new DecimalFormat("#,###");
+
+    @Override
+    public boolean setup() {
+        Optional<Service<EconomyProvider>> serviceOpt = ServiceRegistry.INSTANCE.serviceFor(
+                EconomyProvider.class);
+        if (serviceOpt.isPresent()) {
+            provider = serviceOpt.get().get();
+            return true;
+        } else {
+            return false;
+        }
+    }
+
+    @Override
+    public boolean canRegister() {
+        return provider != null;
+    }
+
+    @Override
+    public @Nullable String onRequest(
+            @Nullable OfflinePlayer player, @NotNull String param
+    ) {
+        if (provider != null && player != null) {
+            if (param.equalsIgnoreCase("balance")) {
+                return format.format(EconomySubscriber
+                        .<Boolean>asFuture(s -> provider.hasPlayerAccount(player.getUniqueId(), s))
+                        .thenCompose(val -> {
+                            if (val) {
+                                return EconomySubscriber.<PlayerAccount>asFuture(s -> provider.retrievePlayerAccount(
+                                        player.getUniqueId(),
+                                        s
+                                ));
+                            } else {
+                                return EconomySubscriber.<PlayerAccount>asFuture(s -> provider.createPlayerAccount(
+                                        player.getUniqueId(),
+                                        s
+                                ));
+                            }
+                        })
+                        .thenCompose(account -> EconomySubscriber.<BigDecimal>asFuture(s -> account.retrieveBalance(
+                                provider.getPrimaryCurrency(),
+                                s
+                        )))
+                        .join()
+                        .doubleValue());
+            }
+        }
+        return null;
+    }
+
+}
--- a/platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/PAPIExpansion.java
+++ b/platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/PAPIExpansion.java
@@ -1,0 +1,63 @@
+/*
+ * This file is/was part of Treasury. To read more information about Treasury such as its licensing, see <https://github.com/lokka30/Treasury>.
+ */
+
+package me.lokka30.treasury.plugin.bukkit.hooks;
+
+import me.clip.placeholderapi.expansion.PlaceholderExpansion;
+import me.lokka30.treasury.plugin.bukkit.TreasuryBukkit;
+import me.lokka30.treasury.plugin.core.TreasuryPlugin;
+import org.bukkit.OfflinePlayer;
+import org.jetbrains.annotations.NotNull;
+
+public class PAPIExpansion extends PlaceholderExpansion {
+
+    private final String author;
+    private TreasuryPAPIHook economy;
+
+    public PAPIExpansion(TreasuryBukkit plugin) {
+        this.author = String.join(", ", plugin.getDescription().getAuthors());
+        this.economy = new EconomyHook();
+    }
+
+    @Override
+    public boolean register() {
+        if (!economy.setup()) {
+            economy = null;
+        }
+        if (economy != null) {
+            return super.register();
+        }
+        return false;
+    }
+
+    @Override
+    public @NotNull String getIdentifier() {
+        return "treasury";
+    }
+
+    @Override
+    public @NotNull String getAuthor() {
+        return author;
+    }
+
+    @Override
+    public @NotNull String getVersion() {
+        return TreasuryPlugin.getInstance().getVersion().toString();
+    }
+
+    @Override
+    public boolean canRegister() {
+        return economy != null && economy.canRegister();
+    }
+
+    @Override
+    public String onRequest(OfflinePlayer player, @NotNull String param) {
+        if (economy != null && param.startsWith("economy_")) {
+            return economy.onRequest(player, param.replace("economy_", ""));
+        }
+
+        return null;
+    }
+
+}
--- a/platform/bukkit/plugin/pom.xml
+++ b/platform/bukkit/plugin/pom.xml
@@ -78,6 +78,10 @@
             <name>Minecraft Libraries</name>
             <url>https://libraries.minecraft.net</url>
         </repository>
+        <repository>
+            <id>placeholderapi</id>
+            <url>https://repo.extendedclip.com/content/repositories/placeholderapi/</url>
+        </repository>
     </repositories>
 
     <dependencies>
@@ -137,6 +141,12 @@
             <artifactId>bstats-bukkit</artifactId>
             <version>${bstats.version}</version>
             <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>me.clip</groupId>
+            <artifactId>placeholderapi</artifactId>
+            <version>${placeholderapi.version}</version>
+            <scope>provided</scope>
         </dependency>
     </dependencies>
 
--- a/platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/TreasuryBukkit.java
+++ b/platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/TreasuryBukkit.java
@@ -17,6 +17,7 @@ import me.lokka30.treasury.api.economy.misc.OptionalEconomyApiFeature;
 import me.lokka30.treasury.plugin.bukkit.command.TreasuryCommand;
 import me.lokka30.treasury.plugin.bukkit.event.bukkit2treasury.B2TEventMigrator;
 import me.lokka30.treasury.plugin.bukkit.event.treasury2bukkit.T2BEventMigrator;
+import me.lokka30.treasury.plugin.bukkit.hooks.PAPIExpansion;
 import me.lokka30.treasury.plugin.bukkit.services.ServiceMigrationManager;
 import me.lokka30.treasury.plugin.bukkit.services.bukkit2treasury.B2TServiceMigrator;
 import me.lokka30.treasury.plugin.bukkit.services.treasury2bukkit.T2BServiceMigrator;
@@ -86,6 +87,10 @@ public class TreasuryBukkit extends JavaPlugin {
 
         getServer().getPluginManager().registerEvents(new B2TServiceMigrator(), this);
         new T2BServiceMigrator(this).registerListeners();
+
+        if (getServer().getPluginManager().isPluginEnabled("PlaceholderAPI")) {
+            new PAPIExpansion(this).register();
+        }
 
         if (BukkitVendor.isPaper()) {
             PaperEnhancements.enhance(this);
--- a/platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/TreasuryPAPIHook.java
+++ b/platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/TreasuryPAPIHook.java
@@ -1,0 +1,19 @@
+/*
+ * This file is/was part of Treasury. To read more information about Treasury such as its licensing, see <https://github.com/lokka30/Treasury>.
+ */
+
+package me.lokka30.treasury.plugin.bukkit.hooks;
+
+import org.bukkit.OfflinePlayer;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public interface TreasuryPAPIHook {
+
+    boolean setup();
+
+    boolean canRegister();
+
+    @Nullable
+    String onRequest(@Nullable OfflinePlayer player, @NotNull String param);
+}
--- a/platform/bukkit/plugin/src/main/resources/plugin.yml
+++ b/platform/bukkit/plugin/src/main/resources/plugin.yml
@@ -19,7 +19,7 @@ authors: [ "lokka30", "Jikoo" , "MrIvanPlays" ]
 
 # Dependencies
 depend: [ ]
-softdepend: [ ]
+softdepend: [ "PlaceholderAPI" ]
 
 # General info for the plugin loader
 main: "me.lokka30.treasury.plugin.bukkit.TreasuryBukkit"
--- a/pom.xml
+++ b/pom.xml
@@ -47,6 +47,7 @@
         <bstats.version>3.0.0</bstats.version>
         <annotationconfig.version>2.1.1-SNAPSHOT</annotationconfig.version>
         <mockito.version>4.3.1</mockito.version>
+        <placeholderapi.version>2.11.1</placeholderapi.version>
     </properties>
 
     <licenses>


platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/PAPIExpansion.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @Override
+    public boolean canRegister() {
+        return economy != null && economy.canRegister();
+    }
+
+    @Override
+    public String onRequest(OfflinePlayer player, @NotNull String param) {
+        if (economy != null && param.startsWith("economy_")) {
```
lokka30, line 56: I think we should use a `treasury_eco_` prefix instead, just like the PAPI team did with their Vault expansion.

platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/EconomyHook.java
```diff
@@ -0,0 +1,73 @@
...
+import me.lokka30.treasury.api.economy.account.PlayerAccount;
+import me.lokka30.treasury.api.economy.response.EconomySubscriber;
+import org.bukkit.OfflinePlayer;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class EconomyHook implements TreasuryPAPIHook {
+
+    private EconomyProvider provider;
+    private final DecimalFormat format = new DecimalFormat("#,###");
```
lokka30, line 22: Should we use `#,##0.00`?

MrIvanPlays: It is 0,00 currently. We can change it to 0.00

lokka30: From memory, using the DF I suggested, it will look like this:
* `1.2345` = `1.23`
* `25.3` = `25.30`
* `1234` = `1,234.00`
* `1234567.89` = `1,234,567.89` (unsure if there will be two commas or only one)

Jikoo: > It is 0,00 currently. We can change it to 0.00

The period is just a decimal separator notation and the comma is a grouping separator notation specifically for declaring the format. You can force a specific locale's symbols by specifying the `DecimalFormatSymbols` to be used, but otherwise they will be localized by the formatter for your system locale. 

> From memory, using the DF I suggested, it will look like this:
> * `1234567.89` = `1,234,567.89` (unsure if there will be two commas or only one)

There will be two. The number of characters between the last grouping separator and the decimal separator (or the end of the pattern) is applied to any additional multiples of the grouping separator. A pattern `#,##,###` applied to `100000` should actually yield `100,000` and not `1,00,000`.

lokka30: Unless you don't like it, perhaps we could ask the PAPI team if it's a better decimal format they could use for their Vault one as well

MrIvanPlays:
List of all placeholders added:

# Top balance
These placeholders do not need a player object to function.
`treasury_eco_top_balance`
`treasury_eco_top_balance_{currency}`
`treasury_eco_top_balance_fixed`
`treasury_eco_top_balance_fixed_{currency}`
`treasury_eco_top_balance_formatted`
`treasury_eco_top_balance_formatted_{currency}`
`treasury_eco_top_balance_formatted_{precision}ds`
`treasury_eco_top_balance_commas`
`treasury_eco_top_balance_commas_{currency}`

# Top player
These placeholders do not need a player object to function.
`treasury_eco_top_player`
`treasury_eco_top_player_{currency}`
`treasury_eco_top_player_{currency}_{position}`

# Balance
These placeholders need a player object to function.
`treasury_eco_balance`
`treasury_eco_balance_{currency}`
`treasury_eco_balance_{precision}dp`
`treasury_eco_balance`
`treasury_eco_balance_{currency}`
`treasury_eco_balance_{precision}dp`

MrIvanPlays:
top_rank means we're retrieving their top position. I kept names similar to the vault expansion. And no dp is not a typo

lokka30:
> top_rank means we're retrieving their top position. I kept names similar to the vault expansion. And no dp is not a typo

when no currency is specified, does it assume the primary one?

MrIvanPlays:
Yes

platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/EconomyHook.java
lokka30: Even though the proposed DecimalFormat is better, let's keep it the way PAPI has it

platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/EconomyHook.java
```diff
@@ -0,0 +1,331 @@
...
+    private Baltop baltop;
+
+    public EconomyHook(PAPIExpansion expansion, TreasuryBukkit plugin) {
+        this.expansion = expansion;
+        this.plugin = plugin;
+        this.k = expansion.getString("formatting.thousands", "k");
+        this.m = expansion.getString("formatting.millions", "m");
+        this.b = expansion.getString("formatting.billions", "b");
+        this.t = expansion.getString("formatting.trillions", "t");
+        this.q = expansion.getString("formatting.quadrillions", "q");
```
lokka30, line 38: aren't M, B, T and Q all meant to be capitalized?

platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/EconomyHook.java
```diff
@@ -0,0 +1,331 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class EconomyHook implements TreasuryPAPIHook {
+
+    private EconomyProvider provider;
+    private final DecimalFormat format = new DecimalFormat("#,###");
+    private final PAPIExpansion expansion;
+    private final TreasuryBukkit plugin;
+    private final String k, m, b, t, q;
```
Jikoo, line 28: While I appreciate that this expression of variables is concise, it's generally identified as a code smell due to a lack of readability.

platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/EconomyHook.java
```diff
@@ -0,0 +1,331 @@
...
+        }
+    }
+
+    @Override
+    public boolean canRegister() {
+        return provider != null;
+    }
+
+    @Override
+    public @Nullable String onRequest(
```
Jikoo, line 66: This is pretty unreadable/unmaintainable. If you compare to PAPI's vault expansion, they leverage precompiled Patterns to keep the code clean. IMO we should either create helper methods or more exactly copy PAPI's implementation (though this does then lead to having a modified GPL-licensed file in this MIT project - it may be easier to make a separate module to avoid licensing hell, though theoretically by declaring the particular file GPL and being open source we are already compliant).

Based on a comparison this is already a modified copy of PAPI's Vault expansion, so we may as well use their higher quality version and modify their patterns and helper methods to support Treasury's multi-currency system.

platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/EconomyHook.java
```diff
@@ -0,0 +1,331 @@
...
+            @Nullable OfflinePlayer player, @NotNull String param
+    ) {
+        if (provider != null) {
+            if (!baltop.isEnabled() && param.startsWith("top_")) {
+                return (param.startsWith("top_balance")) ? "0" : "";
+            }
+
+            if (param.contains("top_balance")) {
+                if (param.startsWith("top_balance_fixed")) {
+                    String currencyId = param.replace("top_balance_fixed_", "");
```
Jikoo, line 76: This results in a runaround for `top_balance_fixed` where a regular expression would not. Same for all other entries that accept parameters but do not require them.

Jikoo (changes requested):

platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/EconomyHook.java
Jikoo: They are, and it makes it even more clear that this is a modified copy of PAPI's Vault expansion. All the file headers should contain GPL notice.
https://github.com/PlaceholderAPI/Vault-Expansion/blob/63ddf694fdc1c99940d49e513469cb5780cfba24/src/main/java/com/extendedclip/papi/expansion/vault/VaultExpansion.java#L58-L62
https://github.com/PlaceholderAPI/Vault-Expansion/blob/63ddf694fdc1c99940d49e513469cb5780cfba24/src/main/java/com/extendedclip/papi/expansion/vault/VaultEcoHook.java#L73-L77

lokka30: This is not a crucial issue, since the APIs remains intact. However, I am not pleased that we need to borrow code, especially that from a more restrictive license. I think this is something important we need to address before merging.

platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/EconomyHook.java
MrIvanPlays: what shall i change here?

platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/EconomyHook.java
MrIvanPlays: i have no idea on how to use patterns.

platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/EconomyHook.java
MrIvanPlays: i have no idea on how to use patterns.

platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/EconomyHook.java
Jikoo: Variables should be expressed 1 per line. It's not hugely important, this can definitely be ignored, but adhering to blanket rules makes it easier to justify enforcing elsewhere where it's more important.

https://wiki.sei.cmu.edu/confluence/display/java/DCL52-J.+Do+not+declare+more+than+one+variable+per+declaration

platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/EconomyHook.java
Jikoo: It's used in the plugin side that API users won't depend on, it should be fine imo. It does get a bit messy, but if we're using the same placeholder design we probably shouldn't bother reinventing the wheel.

platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/EconomyHook.java
Jikoo: Do you need an explanation of the existing Pattern usage/regular expressions in PAPI? They're pretty simple. Overall would highly recommend learning the basics of how to write regex - it's tremendously useful for searches, find and replace, etc. I will admit some more complicated patterns are absolutely arcane bullhonkey though.

MrIvanPlays: I tried learning patterns and I can't get anything to work with them. Googling isn't my power either.

MrIvanPlays: also, i don't know how can I make this more readable.

MrIvanPlays:
@Jikoo anything you want to improve im all for you pushing to this branch.

Jikoo:
Should have time for this today.

platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/EconomyHook.java
```diff
@@ -0,0 +1,331 @@
...
+                    if (topBalanceFixed == null) {
+                        return "0";
+                    } else {
+                        return String.valueOf(topBalanceFixed.longValue());
+                    }
+                }
+                if (param.startsWith("top_balance_formatted")) {
+                    Locale locale = Locale.ENGLISH;
+                    if (player != null && player.isOnline()) {
+                        locale = Locale.forLanguageTag(player.getPlayer().getLocale());
```
Jikoo, line 90: Note: afaik the client's language is not ever a real language tag - it uses `en_GB` etc. instead of `en-GB`, and allows pretty much anything via custom resource pack. The default client also includes several joke languages such as Pirate (`en_PT`) that may also cause trouble.

Jikoo: I elected to bandaid this by replacing `_` with `-` but we may wish to split it out to a more robust utility method.

Jikoo:
I still have a decent amount to go but I'm unfortunately out of time for now.

Big things:

* Test the regular expression
  * My aim was to have a single expression that will easily capture named groups so it's very clear (at least in the Pattern/Matcher usage - regex is horrific to read) exactly what is being pulled from the parsed string where. My goal in using lookaheads was specifically to prevent invalid formats like `top_balance_formatted_1dpapa` from registering as 1 decimal place, but I have not yet tested it. Theoretically it might work as intended, but it might capture the underscores from the following groups and prevent them from parsing.
* Address the player side of baltop

platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/EconomyHook.java
Jikoo: Honestly the Pattern javadoc is largely all you need for a resource. It's a lot of reading, and it can get messy in a hurry, but I can't even begin to count the number of times using a simple regular expression has allowed me to digest or convert a pile of useless data into a more readable or functional form.

Re: readability in general, mostly the economy data parsing is/was one gigantic heap, high indentation and very run-on with easy breaking points for different parsing modes.

platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/Baltop.java
```diff
@@ -21,14 +23,16 @@
...
 public class Baltop extends BukkitRunnable {
 
     private final boolean enabled;
-    private final int topSize, taskDelay;
+    private final int topSize;
+    private final int taskDelay;
     private final EconomyProvider provider;
 
-    private Multimap<String, TopPlayer> baltop;
+    private final Multimap<String, TopPlayer> baltop;
```
MrIvanPlays: wouldn't making it final require a lock?

Jikoo: No, it's only initialized once. A final value is still mutable, it is just not reassignable. It already was not reassignable in the existing design, this was just a clarification of that.

MrIvanPlays: no the thing is that i know final maps require either a lock to get accessed or a `synchronized` block

Jikoo: The `final` keyword exclusively prevents reassignment of a reference. It does nothing more and nothing less. It has nothing to do with concurrency. While the maps and collections used should likely be concurrent due to the fact that they're populated on a different thread, the keyword being present does not affect that.

MrIvanPlays:
uh oh - started writing some tests
image
`testTopBalanceFail2` looks like I did it accidentally right but the test should fail but it passes.

MrIvanPlays:
committed pattern according to both intellij and java is wrong ; intellij suggests some brackets. i am unsure what shall i do so im not going to commit the test.

MrIvanPlays:
@Jikoo this doesn't fix it

Jikoo:
If you do want to write tests, the part I'm shaky on is the lookahead, `(?=(_|$))`, which should require either end of input or an underscore as the next character but not include it in the optional group. Removing that should cause the other sections to succeed at least, though the mentioned `10dpwhatever` case will fail.

> @Jikoo this doesn't fix it

What does the parser suggest? Frankly, this is why I marked it as a draft - I'm good with basic regex, this is more advanced and I usually take a bit of trial and error. I'll get to it, but currently the parenthesis are actually balanced (whoops) and the groups are theoretically where they should be.

Either way, running late and need to be heading away from my PC for a while, I'll be around on mobile but won't be able to fix any of my mistakes beyond basic edits.

MrIvanPlays:
i found what's wrong with the succeeding tests. will try to fix the pattern and perhaps transform more code into patterns.

Jikoo:
should also note you should remove the `treasury_eco_` prefix from all those test strings.

MrIvanPlays:
@Jikoo added some clarification on what you need to fix

lokka30:
Want me to work on anything? I am terrible at Regex, so unfortunately I am unable to do anything about the troubles with that.

MrIvanPlays:
No @lokka30

Jikoo:
I think the solution is to make the currency group reluctant - it unfortunately needs to capture any and everything because currency names are unrestricted. Theoretically an economy provider could add a currency with the ID `$%^&*`, which the expansion would then not be able to convert. That was why I put it last. Will play around with it today.

Jikoo:
Note that a failure to parse a parameter will generally result all following items being lumped into currency. This is expected behavior, i.e. `treasury_eco_top_balance_fixef_3_dollar` should be looking for the currency `fixef_3_dollar`. There's not much we can do about this without demanding currency validation on each and every query with a null balance. Thoughts are welcome on this.

Theoretically in a production environment null balances would largely never occur so it shouldn't have a performance impact, and it would make setup more user-friendly.
We should also consider making the base expansion cacheable and having part of the cache be the currency lookup result. That enables reloading it via PAPI I believe, which while not super obvious is definitely a good way to avoid repeated potential async lookups on the main thread.

MrIvanPlays:
@Jikoo I've cleared up the TODOs, the only thing TODO is the adding of more multiples in `fixMoney`. I don't think it is necessary, but I wouldn't mind if you want to add more.

MrIvanPlays:
Updated list of placeholders:

# No player object needed

## Top balance
`treasury_eco_top_balance`
`treasury_eco_top_balance_{currency}`
`treasury_eco_top_balance_{rank}`
`treasury_eco_top_balance_fixed`
`treasury_eco_top_balance_fixed_{currency}`
`treasury_eco_top_balance_fixed_{rank}`
`treasury_eco_top_balance_fixed_{rank}_{currency}`
`treasury_eco_top_balance_formatted`
`treasury_eco_top_balance_formatted_{currency}`
`treasury_eco_top_balance_formatted_{rank}`
`treasury_eco_top_balance_formatted_{precision}dp`
`treasury_eco_top_balance_formatted_{precision}dp_{rank}`
`treasury_eco_top_balance_formatted_{precision}dp_{currency}`
`treasury_eco_top_balance_formatted_{precision}dp_{rank}_{currency}`
`treasury_eco_top_balance_commas`
`treasury_eco_top_balance_commas_{currency}`
`treasury_eco_top_balance_commas_{rank}`
`treasury_eco_top_balance_commas_{rank}_{currency}`

## Top player
`treasury_eco_top_player`
`treasury_eco_top_player_{rank}`
`treasury_eco_top_player_{currency}`
`treasury_eco_top_player_{rank}_{currency}`

# Player object needed

## Top rank
`treasury_eco_top_rank`
`treasury_eco_top_rank_{currency}`

## Balance
`treasury_eco_balance`
`treasury_eco_balance_{currency}`
`treasury_eco_balance_fixed`
`treasury_eco_balance_fixed_{currency}`
`treasury_eco_balance_formatted`
`treasury_eco_balance_formatted_{currency}`
`treasury_eco_balance_formatted_{precision}dp`
`treasury_eco_balance_formatted_{precision}dp_{currency}`
`treasury_eco_balance_commas`
`treasury_eco_balance_commas_{currency}`

Jikoo:
I'd still like to retouch the balance placeholders and other area fetching currencies to cache the responses. Ideally the Extension should implement Cacheable so PAPI knows it can be cleared. I feel like while most economy providers are likely to keep Currency data in memory, on the off chance that they don't I don't want us to be causing lag while it loads over and over. I can get to that tomorrow.

MrIvanPlays:
@Jikoo what happened?

Jikoo:
Life! I've in fact just loaded up Treasury in my IDE now.

Jikoo:
And almost immediately got called away for things after that. Any rate, finally had a look through. It seems like you addressed most things I wanted to touch.

The only thing that still kinda bothers me is that Treasury is loading its expansion on enable. At a first glance this is fine, but the expansion is inflexible and does not allow on-the-fly swapping of the economy in play, which means that since there won't be one when Treasury loads (as Treasury will load before dependents) the expansion will theoretically never work. The PAPI Vault expansion is similarly limited, but at least loads later so it can at least connect with a single economy. Guess that and Cacheable will be my additions today 👍

MrIvanPlays:
i am not really against having a way of hotswapping providers, but did i something about why such system has been implemented?

Jikoo:
The current system does not support hotswapping providers. It also enables and fully completes load for `<current provider>` when Treasury enables. However, because Treasury will enable before any providers (because the providers require Treasury to be loaded to provide via it) there will never be a provider enabled and Treasury's provider placeholders will never work.

PAPI-Vault also does not support hotswapping providers, but it also loads later, after the providers have been loaded, and functions as a result.

MrIvanPlays:
no no my question was about `/treasury economy migrate`. but you brought a point ; perhaps we can register when ServiceRegisteredEvent gets called for an economy provider? /shrug . im gonna leave the decision for that up to you.

Jikoo:
`/treasury economy migrate` exists largely for when a server decides to swap to another Treasury implementation. It won't see heavy use, but I definitely had cause to swap core plugins like economies when I managed servers before. Migration commands generally are on a per-plugin basis and may not exist at all or support certain features. Treasury including it out of the box makes any change easy. Whether you're swapping to a SQL economy provided by the same plugin that manages your flatfile economy or an entirely different plugin, you just tell Treasury to convert all the data and you're done, no need to worry if X economy can read all the data from and import from Y economy.

Imo it's important that the idea of migration be included in Treasury so that Treasury always ensures all implementations expose every detail (such as all existing player accounts' IDs) that are needed to completely migrate all information, a bit of a self-check. It also saves needing the same migrator code in every single economy implementation.

MrIvanPlays:
@Jikoo do you want me to setup a cache for the balances or you will do that?

Jikoo:
> @Jikoo do you want me to setup a cache for the balances or you will do that?

Sorry, not super clear on what you mean here. Doesn't BalTop cache balances? I figure since it runs async it should be fine with whatever update frequency.

As far as the PR goes I think that's it for me, ready to convert back from a draft whenever you are - the only thing I'm still tempted to do is repackage/rename to be slightly more descriptive, but those are largely just personal nitpicks, the existing names work.

* Package `me.lokka30.treasury.plugin.bukkit.hooks` -> `...hooks.papi` or `...hooks.placeholders`: If we later hook other plugins there will be a lot of clutter.
* Class `PAPIExpansion` -> `TreasuryExpansion`: It's a PlaceholderExpansion for Treasury. The concept of an expansion is from PAPI, I think that's enough of a reference.
* Class `TreasuryPAPIHook` -> `TreasuryHook`: Mostly just a nitpick based on liking Google style, acronyms should either not be in class names or should follow casing conventions for words, i.e. `TreasuryPapiHook`. This is also dependent on repackaging, I would not want a more generic name in the same generic package.

MrIvanPlays:
> /e: Oh, right, the balance placeholder that actively queries, shoot. That's tough, not sure how I want to approach that. If you have ideas, please feel free.

got it right the second time. 
what about having a map cache that resets every 10 minutes?

MrIvanPlays:
@Jikoo i suggest you to have a read: https://github.com/lokka30/Treasury/wiki/Treasury-Events-API

Jikoo:
10 minutes seems like a long time for something that would probably be actively displayed on some servers - balance is a classic thing to see in a scoreboard on the side, particularly if the balance encompasses "tokens" or some other specialty currency.

My only idea so far is a class that stores a value, has a method to query it, and stores the last update timestamp:
* If last update is over X age, trigger update and provide resulting value
* If last update is under X age and over Y age, trigger update async and provide current value
* If last update is under Y age, provide current value

Jikoo:
Hah, whoops, my bad. What I get for rushing through that bit 🤦

MrIvanPlays:
my idea for caching is pretty simple: have some type of cache for x amount of time, on request check if we have something and if we do - return that, if we don't - populate and return the value we received.

lokka30:
Did you guys consider just changing the cached economy provider on ServiceRegisterEvent?

MrIvanPlays:
We should change on unregister too, because the provider we were using may not exist anymore.

MrIvanPlays:
actually got it working, whole explanations are wrong. expect commit soon

https://user-images.githubusercontent.com/37958105/156437008-e0780f12-a379-49e0-962f-0731a14de9c8.mp4

platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/HookRegistrator.java
```diff
@@ -0,0 +1,45 @@
...
+    public void onEnable(PluginEnableEvent event) {
+        Plugin plugin = event.getPlugin();
+        if (!hooks.containsKey(plugin.getName())) {
+            return;
+        }
+        for (Hook hook : hooks.values()) {
+            if (hook.register(treasuryBukkit)) {
+                treasuryBukkit
+                        .getLogger()
+                        .info("Treasury " + plugin.getName() + " hook registered successfully");
```
MrIvanPlays, line 40: perhaps we shall remove "Treasury" from here because currently for PAPI it is 
`[Treasury] Treasury PlaceholderAPI hook registered successfully`

Jikoo (approved):
Looks good to me 👍

PR #159, opened 2022-02-18
Update javadoc

lokka30:
When this PR is ready for review, it will be marked as such. If it's a draft, don't merge it 😉 

***

To-do:
- [ ] Inform developers of the 'best practices' when using `EconomyProvider…#retrievePlayerAccount` and all other retrieve methods too.
- [ ] Update documentation for using subscribers and the completablefuture conversion. Neither of them check `!has account -> create account`.
- [ ] ...more? may as well cover it in this PR.


lokka30:
> i think that we need contracts for every retrieve method.

Treasury definitely needs clear "do's and dont's" on its APIs. We can't allow a Vault situation to occur

lokka30:
I'd rather implement this later

PR #164, opened 2022-03-04
First preparations for event calling

yannicklamprecht:
Implements Event calling.

# TODOS:

- [x] Wrap Subscriber to allow post actions events
- [ ] Call pre action events and handle cancel status


yannicklamprecht:
A lot of events aren't defined yet, so it's in a blocked state.

Probably Missing Events?!:

- [ ] AccountNameChangeEvent
- [ ] AccountNameChangedEvent
- [ ] AccountRetrieveBalanceEvent
- [ ] AccountRetrievedBalanceEvent
- [ ] AccountBalanceModifyEvent
- [ ] AccountBalanceModifiedEvent
- [ ] AccountDoTransactionEvent
- [ ] AccountDidTransactionEvent
- [ ] AccountDeleteEvent
- [ ] AccountDeletedEvent
- [ ] AccountHistoryRetrieveEvent
- [ ] AccountHistoryRetrievedEvent
- [ ] AccountMemberIdsRetrievedEvent
- [ ] AccountIsMemberEvent
- [ ] AccountModifyPermissionEvent
- [ ] AccountModifiedPermissionEvent
- [ ] AccountPermissionsRetrieveEvent
- [ ] AccountPermissionsRetrievedEvent
- [ ] AccountPermissionCheckEvent
- [ ] AccountPermissionCheckedEvent

- [ ] PlayerAccountBalanceResetEvent

lokka30:
Thank you very much! I have a few questions:

- Are you sure this PR is no longer a draft?
- How is the EventEconomyProvider utilized? I don't see any entry points for it.

And here are my opinions on each event you listed above:

<details>
<summary>[View Opinions...]</summary>

- I would prefer using the word `Update` than `Change` and `Modify` in the event names.
- Unsure if I prefer present/post-tense words (e.g. `do-transaction` and `did-transaction`) or using prefixes like `Pre` and `Post` (e.g. `pre-transaction` and `post-transaction`).
- I'm unsure about any of these events being modifiable / cancellable. It could break plugins by cancelling them or changing their state.

Yes:
- `AccountBalanceModifyEvent`
- `AccountBalanceModifiedEvent`
- `AccountDoTransactionEvent`
- `AccountDidTransactionEvent`
- `AccountDeleteEvent`
- `AccountDeletedEvent`
- `AccountModifyPermissionEvent`
- `AccountModifiedPermissionEvent`

Maybe:
- `AccountNameChangeEvent`
- `AccountNameChangedEvent`
- `AccountRetrieveBalanceEvent`
- `AccountRetrievedBalanceEvent`
- `AccountHistoryRetrieveEvent`
- `AccountHistoryRetrievedEvent`
- `AccountMemberIdsRetrievedEvent`
- `AccountIsMemberEvent`
- `AccountAddMemberEvent` and `AccountRemoveMemberEvent`
    - added these two to the list, thought they were missing from the original
- `AccountPermissionsRetrieveEvent`
- `AccountPermissionsRetrievedEvent`

Probably no:
- `PlayerAccountBalanceResetEvent`
- `AccountPermissionCheckEvent`
- `AccountPermissionCheckedEvent`

</details>

MrIvanPlays:
as i see it, this api you will introduce will be a wrapper of the implemented api by economy providers, am i right? if i am, i don't really like this approach, but if others are agreeing i won't stop from merging.

lokka30:
> as i see it, this api you will introduce will be a wrapper of the implemented api by economy providers, am i right? if i am, i don't really like this approach, but if others are agreeing i won't stop from merging.

I have the same opinion if such is the case.

yannicklamprecht:
Feel free to close. 👍🏼 Yeah it's a wrapper.
The alternative could be the Template Pattern. https://www.tutorialspoint.com/design_pattern/template_pattern.htm

lokka30:
> Feel free to close. 👍🏼 Yeah it's a wrapper. The alternative could be the Template Pattern. https://www.tutorialspoint.com/design_pattern/template_pattern.htm

The template pattern sounds like a brilliant alternative! @MrIvanPlays do you think that route would be viable for this?

MrIvanPlays:
that's gonna rock the boat too much. in which way we are obliged to call them? fgs they're only 2 events to call! the economy provider can do this easily.

lokka30:
> that's gonna rock the boat too much. in which way we are obliged to call them? fgs they're only 2 events to call! the economy provider can do this easily.

Although I don't think it is at all *necessary*, if we do go with the template pattern then we can add several other events (see my original reply with the events list in a dropdown).

However I am currently more orientated to keep things the way it is.

lokka30:
Regardless of the PR being closed I am very grateful for your contribution @yannicklamprecht. Thank you very much. :)

PR #180, opened 2022-05-03
Add future default methods

MrIvanPlays:
As discussed [here](https://discord.com/channels/752310043214479462/921225503531204658/970524975993856061) on the [Discord server](https://www.discord.io/arcaneplugins).

***

To-do:

- [ ] Resolve remaining comments.
- [ ] ?..

***

Basically, this will be the 2nd and last breaking change we will do in v2 - removing subscribers - what this is adding is default `CompletableFuture` implementations.

### DIFF
--- a/api/src/main/java/me/lokka30/treasury/api/common/misc/FutureHelper.java
+++ b/api/src/main/java/me/lokka30/treasury/api/common/misc/FutureHelper.java
@@ -1,0 +1,54 @@
+package me.lokka30.treasury.api.common.misc;
+
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.Objects;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Function;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * A utility class to help with {@link CompletableFuture} stuff
+ *
+ * @author MrIvanPlays
+ * @since TODO
+ */
+public class FutureHelper {
+
+    /**
+     * Map, join, filter - this is what the name of the method is.
+     * Maps all the futures holding data type A to return a collection of data type B
+     * via the specified {@code mapper} {@link Function} and filters them via the specified
+     * filter function.
+     * <br><b>WARNING: This is a thread unsafe method. ASYNCHRONOUS CALL MANDATORY!!!!</b>
+     *
+     * @param mapper  mapper
+     * @param futures futures of data type A
+     * @param <A>     type parameter A
+     * @param <B>     type parameter B
+     * @return future with collection of data type B
+     */
+    @NotNull
+    public static <A, B> CompletableFuture<Collection<B>> mjf(
+            @Nullable Function<A, CompletableFuture<Boolean>> filter,
+            @NotNull Function<A, B> mapper,
+            @NotNull Collection<CompletableFuture<A>> futures
+    ) {
+        Objects.requireNonNull(mapper, "mapper");
+        Objects.requireNonNull(futures, "futures");
+        Collection<B> ret = new HashSet<>();
+        for (CompletableFuture<A> future : futures) {
+            if (filter != null) {
+                A a = future.join();
+                if (filter.apply(a).join()) {
+                    ret.add(mapper.apply(a));
+                }
+            } else {
+                ret.add(mapper.apply(future.join()));
+            }
+        }
+        return CompletableFuture.completedFuture(ret);
+    }
+
+}
--- a/api/src/main/java/me/lokka30/treasury/api/common/misc/FutureHelper.java
+++ b/api/src/main/java/me/lokka30/treasury/api/common/misc/FutureHelper.java
@@ -1,12 +1,14 @@
 package me.lokka30.treasury.api.common.misc;
 
+import java.util.AbstractMap;
+import java.util.ArrayList;
 import java.util.Collection;
-import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.concurrent.CompletableFuture;
 import java.util.function.Function;
 import org.jetbrains.annotations.NotNull;
-import org.jetbrains.annotations.Nullable;
 
 /**
  * A utility class to help with {@link CompletableFuture} stuff
@@ -17,11 +19,10 @@ import org.jetbrains.annotations.Nullable;
 public class FutureHelper {
 
     /**
-     * Map, join, filter - this is what the name of the method is.
+     * Maps, joins, and filters futures.
      * Maps all the futures holding data type A to return a collection of data type B
      * via the specified {@code mapper} {@link Function} and filters them via the specified
      * filter function.
-     * <br><b>WARNING: This is a thread unsafe method. ASYNCHRONOUS CALL MANDATORY!!!!</b>
      *
      * @param mapper  mapper
      * @param futures futures of data type A
@@ -30,25 +31,37 @@ public class FutureHelper {
      * @return future with collection of data type B
      */
     @NotNull
-    public static <A, B> CompletableFuture<Collection<B>> mjf(
-            @Nullable Function<A, CompletableFuture<Boolean>> filter,
+    public static <A, B> CompletableFuture<Collection<B>> mapJoinFilter(
+            @NotNull Function<A, CompletableFuture<Boolean>> filter,
             @NotNull Function<A, B> mapper,
             @NotNull Collection<CompletableFuture<A>> futures
     ) {
+        Objects.requireNonNull(filter, "filter");
         Objects.requireNonNull(mapper, "mapper");
         Objects.requireNonNull(futures, "futures");
-        Collection<B> ret = new HashSet<>();
+
+        List<CompletableFuture<Map.Entry<A, Boolean>>> filteredFutures =
+                new ArrayList<>(futures.size());
         for (CompletableFuture<A> future : futures) {
-            if (filter != null) {
-                A a = future.join();
-                if (filter.apply(a).join()) {
-                    ret.add(mapper.apply(a));
-                }
-            } else {
-                ret.add(mapper.apply(future.join()));
-            }
+            filteredFutures.add(future.thenCompose((a) -> {
+                return filter.apply(a)
+                        .thenApply((allowed) -> new AbstractMap.SimpleImmutableEntry<>(a, allowed));
+            }));
         }
-        return CompletableFuture.completedFuture(ret);
+        return CompletableFuture.allOf(
+                filteredFutures.toArray(new CompletableFuture[0])
+        ).thenApply((ignore) -> {
+
+            List<B> results = new ArrayList<>(filteredFutures.size());
+            for (CompletableFuture<Map.Entry<A, Boolean>> filteredFuture : filteredFutures) {
+                // By now, all futures are complete -- join() will not block
+                Map.Entry<A, Boolean> entry = filteredFuture.join();
+                if (entry.getValue()) {
+                    results.add(mapper.apply(entry.getKey()));
+                }
+            }
+            return results;
+        });
     }
 
 }
--- a/api/src/main/java/me/lokka30/treasury/api/common/misc/FutureHelper.java
+++ b/api/src/main/java/me/lokka30/treasury/api/common/misc/FutureHelper.java
@@ -12,7 +12,7 @@ import org.jetbrains.annotations.Nullable;
  * A utility class to help with {@link CompletableFuture} stuff
  *
  * @author MrIvanPlays
- * @since TODO
+ * @since 1.2.1
  */
 public class FutureHelper {
 
--- a/api/src/main/java/me/lokka30/treasury/api/economy/EconomyProvider.java
+++ b/api/src/main/java/me/lokka30/treasury/api/economy/EconomyProvider.java
@@ -5,12 +5,14 @@
 package me.lokka30.treasury.api.economy;
 
 import java.util.Collection;
+import java.util.Collections;
 import java.util.HashSet;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.UUID;
 import java.util.concurrent.CompletableFuture;
+import me.lokka30.treasury.api.common.misc.FutureHelper;
 import me.lokka30.treasury.api.common.misc.TriState;
 import me.lokka30.treasury.api.economy.account.Account;
 import me.lokka30.treasury.api.economy.account.AccountPermission;
@@ -37,7 +39,7 @@ public interface EconomyProvider {
 
     /**
      * WARNING: API versions are no longer used as of Treasury v1.1.0. Ignore this method.
-     * 
+     *
      * Get the version of the Treasury API the {@code EconomyProvider} is based on.
      *
      * <p>Warning: The Treasury API version is completely different to any other platform version,
@@ -87,6 +89,14 @@ public interface EconomyProvider {
     );
 
     /**
+     * {@link #hasPlayerAccount(UUID, EconomySubscriber)}
+     */
+    @NotNull
+    default CompletableFuture<Boolean> hasPlayerAccount(@NotNull UUID accountId) {
+        return EconomySubscriber.asFuture(s -> hasPlayerAccount(accountId, s));
+    }
+
+    /**
      * Request an existing {@link PlayerAccount} for a user.
      *
      * @param accountId    the {@link UUID} of the account owner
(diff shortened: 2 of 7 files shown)


lokka30:
clarification: this PR is in preparation for the future breaking change, adding in the methods so economy consumers can benefit from the easier methods without yet breaking economy providers' implementations.

<details>
<summary>Chat log in case anyone is interested for context</summary>

(The chat log has been slightly modified to make it clearer)

***

## @lokka30
Ivan - a thought came across my mind and I'd like to hear your opinion:

Adding overloads on the subscriber methods for CompletableFuture conversions, instead of having to manually wrap it with `Subscriber#asFuture`.

These are just variants of the existing methods that return a `CompletableFuture` instead of using a subscriber parameter 😃  e.g.
```java
// subscriber version
void retrievePlayerAccount(
    @NotNull UUID accountId,
    @NotNull EconomySubscriber<PlayerAccount> subscription
)

// CF version
@NotNull
default CompletableFuture<PlayerAccount> retrievePlayerAccount(
    @NotNull UUID accountId
)

// both methods are present in the same file.
// nothing is removed. this is just a short hand that
// makes using CFs less tedious in Treasury.
```

in theory this will be cleaner to use than wrapping the subscriber method with Subscriber#asFuture.

People whinge about having to write concurrent code, so I am trying to find ways to make it easier for them to write. this is vital if we aim to have any adoption in the community.

## @MrIvanPlays

Yes it will be cleaner I agree on this @lokka30
They will be default methods
like
default Future retrieve(...) {
  Future future = ...
  retrieve(...) 
  return future
Currently on phone
Tommorrow ill see what I can do
We should have used futures from the start but your concerns were 'its hard for newbies' 


## @lokka30 

Subscriptions are a useful feature, though Treasury should have always used CompletableFutures as the standard methods and provide defaulted subscription methods which just use CFs under the  hood (opposite to what we have currently).
Glad to hear you like the idea 👍

Regarding the Subscription/CF thing: I still firmly vote that we should make a breaking change in Treasury v2 where CFs are the standard instead of Subscriptions. It will help us in the long term. This is the last and only breaking change I vote to make to Treasury, assuming we do not repeat the current mistakes in any new APIs.

Another regret I have is releasing Treasury on SpigotMC with a release version (v1) rather than a beta version (v0) - a beta version would have otherwise allowed us to trial APIs for real world use without having to be very scared of the rare breaking change.

I think we have ironed out almost everything with the economy API as is; it seems the only issues remaining are the subscriber/CF scenario, and the lack of documentation (which I will work on in my break).

## @MrIvanPlays 

Documentation is something which can be done at any state

Unfortunately, I have a concern which if we do such breaking changes every major release of treasury then devs will view treasury as a 'can break always' API and push them away

## @lokka30 

I would dread using a library/plugin that breaks their API often - this is something I am currently dealing with for one of my own plugins (which try offer compat for another plugin which constantly breaks their API).

I think we have only made one breaking change so far, which was to remove the API versioning system. I think this CF/Sub scenario is important enough that it deserves a breaking change soon. It will make it much easier for economy providers to be written since they can use subscriptions at their own option. This would be breaking change number 2 and almost certainly the last breaking change we will need to make to the economy API... I sincerely hope.

For existing economy providers, they can either shift their usage of subscriptions to CFs, or they can do this:
- override defaults for subscription methods to not use the default, and rather, use the subscription code they've alreay made
- make the CF methods redirect to the subscription methods they've made
Using CFs will only make provider code cleaner so it's up to them really.

</details>

lokka30 (approved):

lokka30:
@MrNemo64, do you have any opinion on this PR?

MrNemo64:
> @MrNemo64, do you have any opinion on this PR?

Can't look at it today (probably) and I'd like to check the conversarion at discord, but the server/channel seems to be private? Doesn't allow me to read it. I suppose I have to join the server to read it

I'll check it probably tomorrow

lokka30:
> > @MrNemo64, do you have any opinion on this PR?
> 
> Can't look at it today (probably) and I'd like to check the conversarion at discord, ~but the server/channel seems to be private? Doesn't allow me to read it. I suppose I have to join the server to read it~ just needed to join the server
> 
> I'll check it probably tomorrow

Awesome, thank you. No expectations at all, just pinging you in case you're interested. :)

I've included the conversation in my first comment on this PR, but you can also view it in the treasury channel on the Discord.

api/src/main/java/me/lokka30/treasury/api/economy/EconomyProvider.java
```diff
@@ -196,6 +272,38 @@ void createAccount(
...
+                return CompletableFuture.completedFuture(Collections.emptySet());
+            }
+            Collection<CompletableFuture<Account>> futures = new HashSet<>();
+            for (String identifier : identifiers) {
+                futures.add(this.retrieveAccount(identifier));
+            }
+            return FutureHelper.mjf(account -> account.isMember(playerId),
+                    Account::getIdentifier,
+                    futures
+            );
```
MrNemo64, line 303: `FutureHelper#mjf` uses [`CompletableFuture#join()`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#join--) which
> throws an (unchecked) exception if completed exceptionally

but this excepcion is not handled in the `mjf` method or this method, nor is it specifyed in the documentation that it's not handled

api/src/main/java/me/lokka30/treasury/api/economy/EconomyProvider.java
```diff
@@ -206,54 +314,48 @@ void createAccount(
...
+                accounts.add(this.retrieveAccount(identifier));
             }
-            subscription.succeed(ret);
+            return FutureHelper.mjf(
+                    account -> account
+                            .hasPermission(playerId, permissions)
+                            .thenApply(triState -> triState == TriState.TRUE),
+                    Account::getIdentifier,
+                    accounts
+            );
```
MrNemo64, line 358: `FutureHelper#mjf` uses [`CompletableFuture#join()`](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#join--) which
> throws an (unchecked) exception if completed exceptionally

but this excepcion is not handled in the `mjf` method or this method, nor is it specifyed in the documentation that it's not handled

platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/papi/economy/BalTop.java
```diff
@@ -151,9 +148,7 @@ private void handlePlayers(EconomyProvider provider) {
                 continue;
             }
             for (Currency currency : provider.getCurrencies()) {
-                BigDecimal balance = EconomySubscriber
-                        .<BigDecimal>asFuture(s -> account.retrieveBalance(currency, s))
-                        .join();
+                BigDecimal balance = account.retrieveBalance(currency).join();
```
MrNemo64, line 151: `join()` may throw an unchecked exception thats not handled

platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/hooks/papi/economy/BalanceCache.java
```diff
@@ -75,13 +72,12 @@ public void run() {
                 continue;
             }
             for (Currency currency : provider.getCurrencies()) {
-                BigDecimal balance = EconomySubscriber
-                        .<BigDecimal>asFuture(s -> account.retrieveBalance(currency, s))
-                        .join();
+                BigDecimal balance = account.retrieveBalance(currency).join();
```
MrNemo64, line 75: `join()` may throw an unchecked exception thats not handled

MrNemo64 (commented):
Why are we using HashSets as collection of completable futures when CompletableFuture does not override hashCode ?
I have commented some places that I've seen that exceptions may be thrown that doesn't seem to be handled. I may have mist some 

https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/concurrent/CompletableFuture.java

MrIvanPlays:
> Why are we using HashSets as collection of completable futures when CompletableFuture does not override hashCode ?
> 
> I have commented some places that I've seen that exceptions may be thrown that doesn't seem to be handled. I may have mist some 
> 
> 
> 
> https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/util/concurrent/CompletableFuture.java

CompletableFuture does nothing about the result but return it after it receives it. So it doesn't matter which collection type is used. 
Re: exceptions: if an exception is thrown, it will be captured by the CompletableFuture and will call the exception handler, specified by CompletableFuture#exceptionally. It usually isn't our responsibility to handle them, but it looks like the mjf method will need a modification due to the nature of the code that uses it.

Jikoo (commented):
Not a lot to add beyond MrNemo64's concerns. Nice catch of the unsafe operations in the PAPI hook.

I do still somewhat worry about inexperienced developers' implementations (it's easy to make mistakes, as we ourselves displayed in the PAPI hook!) as expressed in https://github.com/lokka30/Treasury/discussions/161#discussioncomment-2204446 and other referenced places, but this really is what I wanted from Treasury in the first place.

MrIvanPlays:
I think I have another idea instead of futures and that is adding a way of combining subscriptions. Something like

retrieveAccount(uuid, Subscription.combine(account -> account.retrieveBalance(...)));

In half an hour I can come up with a better example (currently phone), but what I have in mind is probably the best as it is not breaking change and is much cleaner approach than asFuture and raw subscription.

MrIvanPlays:
@Jikoo @lokka30 

```java
    static <A, B> EconomySubscriber<A> thenCall(
            BiConsumer<A, EconomySubscriber<B>> methodCall, EconomySubscriber<B> subscription
    ) {
        return new EconomySubscriber<A>() {
            @Override
            public void succeed(@NotNull final A a) {
                methodCall.accept(a, subscription);
            }

            @Override
            public void fail(@NotNull final EconomyException exception) {
                subscription.fail(exception);
            }
        };
    }
```
example usage:
```java
        provider.retrievePlayerAccount(
                uuid,
                EconomySubscriber.thenCall(
                        (account, subscription) -> account.retrieveBalance(
                                provider.getPrimaryCurrency(),
                                subscription
                        ),
                        new EconomySubscriber<BigDecimal>() {

                            @Override
                            public void succeed(@NotNull BigDecimal balance) {
                                // ...
                            }

                            @Override
                            public void fail(@NotNull EconomyException exception) {
                                // ...
                            }
                        }
                )
        );
```
what do you think?

api/src/main/java/me/lokka30/treasury/api/common/misc/FutureHelper.java
```diff
@@ -0,0 +1,54 @@
...
+     * <br><b>WARNING: This is a thread unsafe method. ASYNCHRONOUS CALL MANDATORY!!!!</b>
+     *
+     * @param mapper  mapper
+     * @param futures futures of data type A
+     * @param <A>     type parameter A
+     * @param <B>     type parameter B
+     * @return future with collection of data type B
+     */
+    @NotNull
+    public static <A, B> CompletableFuture<Collection<B>> mjf(
```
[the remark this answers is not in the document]
lokka30: Agreed - `mapJoinFilter`, I believe, is short enough that it shouldn't be abbreviated.

lokka30:
> Not a lot to add beyond MrNemo64's concerns. Nice catch of the unsafe operations in the PAPI hook.
> 
> I do still somewhat worry about inexperienced developers' implementations (it's easy to make mistakes, as we ourselves displayed in the PAPI hook!) as expressed in [#161 (comment)](https://github.com/lokka30/Treasury/discussions/161#discussioncomment-2204446) and other referenced places, but this really is what I wanted from Treasury in the first place.

Thank you very much for your feedback. :)

By the way - anyone, even economy providers, can use subscribers if they wish. It's just that we are making CFs the primary approach, and no manual wrapping is required to use them.

This does bring me the thought, should we add a Subscriber->CF conversion utility?

lokka30:
Oops, I posted my reply without seeing Ivan's replies to Jikoo. I'll remove this post once I've read Ivan's replies and updated my post.

MrIvanPlays:
> Side-thought - should we add a Subscriber->CF conversion utility?

We already have such.

> I like how it makes subscription chains look nicer, however, I am concerned about exception handling with the foremost chained calls. Does it run `fail(EconomyException)` on the last chained subscription if any of the previous subscriptions fail?

As you can see by the code I sent:


> ```java
>     static <A, B> EconomySubscriber<A> thenCall(
>             BiConsumer<A, EconomySubscriber<B>> methodCall, EconomySubscriber<B> subscription
>     ) {
>         return new EconomySubscriber<A>() {
>             @Override
>             public void succeed(@NotNull final A a) {
>                 methodCall.accept(a, subscription);
>             }
> 
>             @Override
>             public void fail(@NotNull final EconomyException exception) {
>                 subscription.fail(exception);
>             }
>         };
>     }
> ```


if a subscription fails it calls the taken subscription's failure method.

MrIvanPlays:
> (it's easy to make mistakes, as we ourselves displayed in the PAPI hook!)

`.join` is not used mistakenly in the papi hook - as you know the papi hook's cache loads are being run asynchronously, so calling `.join` has no damage at all. Yes it will throw the exception if the future failed, just like any other thing you do if it fails it throws an exception.

lokka30 (changes requested):
I dismiss my previous approval as I would like MrNemo and Jikoo's requested changes to be satisfied. Otherwise, I'm all ready to merge this in.

Ivan - if you would like me to work on anything in this branch, please let me know.

(I couldn't find where the 'dismiss review' feature was in GH, so I just posted this new review instead.)

MrIvanPlays:
> I dismiss my previous approval as I would like MrNemo and Jikoo's requested changes to be satisfied. Otherwise, I'm all ready to merge this in.
> 
> Ivan - if you would like me to work on anything in this branch, please let me know.
> 
> (I couldn't find where the 'dismiss review' feature was in GH, so I just posted this new review instead.)

what about doing the idea i gave instead of continuing this?

MrNemo64:
> > (it's easy to make mistakes, as we ourselves displayed in the PAPI hook!)
> 
> `.join` is not used mistakenly in the papi hook - as you know the papi hook's cache loads are being run asynchronously, so calling `.join` has no damage at all. Yes it will throw the exception if the future failed, just like any other thing you do if it fails it throws an exception.

Since we are the ones that write the papi hook, we are the ones in charge of handling the exception, arent we? If on the cache the future fails, no one is handling the exception right now

lokka30:
Subscription chaining is a separate feature. I don't see why we can't add subscription chaining in, in addition to this CF defaults change.

lokka30:
Hey @MrIvanPlays, please could you give me a status on this PR? 🙂

MrIvanPlays:
stale

api/src/main/java/me/lokka30/treasury/api/common/misc/FutureHelper.java
```diff
@@ -0,0 +1,54 @@
...
+ * @since 1.2.1
+ */
+public class FutureHelper {
+
+    /**
+     * Map, join, filter - this is what the name of the method is.
+     * Maps all the futures holding data type A to return a collection of data type B
+     * via the specified {@code mapper} {@link Function} and filters them via the specified
+     * filter function.
+     * <br><b>WARNING: This is a thread unsafe method. ASYNCHRONOUS CALL MANDATORY!!!!</b>
```
A248, line 24: Would you like me to fix this? It is entirely possible to turn a collection of futures into a future of a collection, without using `CompletableFuture#join`.

As the method currently stands, it seems non-sensical to return a `CompletableFuture<Collection<B>>` considering this future will always be complete by the time the method returns. It might as well return `Collection<B>`.

MrNemo64: That's actually a good point

yannicklamprecht: Joining misses the purpose of a completable future.

A248: See #196

lokka30:
I've added a to-do list to the original post of this pull request so we can keep track of what changes need to be made regarding the concurrency changes in Treasury v2.

Since work on this project seems to be in sparse bursts, I hope that it'll help us stay on track, rather than trying to reassemble what needs to be done each time.

Please edit this to-do list. [[link to top]](https://github.com/lokka30/Treasury/pull/180#issue-1224415445)

MrIvanPlays:
Will be addressed in a different pr to branch 2.0.0

PR #196, opened 2022-08-26
Make FutureHelper#mapJoinFilter future-friendly

A248:
I changed a few details with this commit:
* The method implementation no longer blocks.
* Renamed to #mapJoinFilter per other suggestions in #180 .
* Changed the filter parameter to be non-null, since all usages were non-null anyway.


MrIvanPlays:
You may as well fix the usages please as now this is thread safe async method.

A248:
All done. If you won't mind, I was unable to resist changing the HashSet to a pre-sized ArrayList.

PR #213, opened 2022-10-30
fix colored message sending in console in bukkit

MrIvanPlays:
perhaps 1.19 broke this but just to make sure im fixing it for all the versions
makes this parse properly: img

### DIFF
--- a/platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/BukkitTreasuryPlugin.java
+++ b/platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/BukkitTreasuryPlugin.java
@@ -162,13 +162,8 @@ public class BukkitTreasuryPlugin extends TreasuryPlugin implements Logger, Sche
 
     @Override
     public void error(String message, Throwable t) {
-        // in that case we're going to strip the colors (if any)
-        // currently this method isn't used for sending messages with colors but just to be on
-        // the right side let's strip if anything is there.
-        if (message.indexOf('&') != -1 || message.indexOf(ChatColor.COLOR_CHAR) != -1) {
-            message = ChatColor.stripColor(message.replace('&', ChatColor.COLOR_CHAR));
-        }
-        plugin.getLogger().log(Level.SEVERE, message, t);
+        // TODO: colorize() will NOT work on Paper
+        plugin.getLogger().log(Level.SEVERE, colorize(message), t);
     }
 
     @Override
--- a/platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/BukkitTreasuryPlugin.java
+++ b/platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/BukkitTreasuryPlugin.java
@@ -129,24 +129,46 @@ public class BukkitTreasuryPlugin extends TreasuryPlugin implements Logger, Sche
 
     @Override
     public void info(String message) {
-        plugin.getServer().getConsoleSender().sendMessage("[Treasury] " + colorize(message));
+        if (BukkitVendor.isPaper()) {
+            plugin.getServer().getConsoleSender().sendMessage("[Treasury] " + colorize(message));
+        } else {
+            plugin.getLogger().info(colorize(message));
+        }
     }
 
     @Override
     public void warn(String message) {
-        plugin.getServer().getConsoleSender().sendMessage(
-                "&e[WARNING] [Treasury] " + colorize(message)
-        );
+        if (BukkitVendor.isPaper()) {
+            plugin
+                    .getServer()
+                    .getConsoleSender()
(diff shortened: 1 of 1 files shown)


Jikoo:
A Paper inconsistency, not 1.19. I think they broke loggers supporting color in their Adventure update, then never added it back in. LGTM 👍

MrIvanPlays:
Colors work but only from the console sender and not from the logger for some weird reason but yeah. In that case i might make that fix apply only for paper but lokka30 has to say whether this is a viable solution or not.

yannicklamprecht:
> A Paper inconsistency, not 1.19. I think they broke loggers supporting color in their Adventure update, then never added it back in. LGTM 👍

That was done intentionally. 
https://github.com/PaperMC/Paper/issues/7596#issuecomment-1073307545

MrIvanPlays:
So you want us to use components? For what when you're gonna send the message to the logger over legacy text? That's just more parsing being done.

yannicklamprecht:
> So you want us to use components? For what when you're gonna send the message to the logger over legacy text? That's just more parsing being done.

Nope, just strip color. The logger intentionally does not interpret colors. You either use the console sender to allow writing legacy colors or just strip the colors.

lokka30:
I believe it is best that we use the Bukkit logger normally (colorized) on Spigot servers, whilst on Paper-based servers, the overridden log method can parse the input into an Adventure component and then output that.

If we prefer the way this PR has implemented it, note that we will need to also put the "&e[WARN] [Treasury]" part inside the colorize method call.

I am on the fence whether I agree with Paper's change for colored logging: yes, I see how loggers should directly output what they were given, however this change has broken plugins and may have been better implemented via an opt-in by individual plugins who desire raw logging. Anyhow, I'll save my ramble since the change has been made and it is off-topic.

platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/BukkitTreasuryPlugin.java
```diff
@@ -129,24 +129,46 @@ private String colorizeHex(@NotNull String message) {
...
 
     @Override
     public void error(String message, Throwable t) {
-        plugin.getLogger().log(Level.SEVERE, colorize(message), t);
+        // in that case we're going to strip the colors (if any)
+        // currently this method isn't used for sending messages with colors but just to be on
+        // the right side let's strip if anything is there.
+        if (message.indexOf('&') != -1 || message.indexOf(ChatColor.COLOR_CHAR) != -1) {
+            message = ChatColor.stripColor(message.replace('&', ChatColor.COLOR_CHAR));
+        }
```
lokka30, line 170: This code will cause `&` characters to be replaced with `§` in messages, regardless if they are used for chat coloring.

Rest of the commit looks fantastic!

Lastly, I'm unfamiliar with the reason why we are stripping colors here, please inform me. :)

lokka30 (changes requested):

platform/bukkit/plugin/src/main/java/me/lokka30/treasury/plugin/bukkit/BukkitTreasuryPlugin.java
lokka30: It may be worth removing any consideration of the ampersand character in this method, since we only parse either non-colored or colorized inputs to this method anyways.

MrIvanPlays: > This code will cause `&` characters to be replaced with `§` in messages, regardless if they are used for chat coloring.

I know. I wanted to avoid parsing colors and then stripping them.

> Lastly, I'm unfamiliar with the reason why we are stripping colors here, please inform me. :)

this is a logging method that also attaches a exception so why shall we have colors here? 

> It may be worth removing any consideration of the ampersand character in this method, since we only parse either non-colored or colorized inputs to this method anyways.

didn't understand what you were trying to say

lokka30: I'll slightly deviate from my original thought and say that after ensuring that no existing usages of this method use colored messages, then there is no need to have the color stripping system.

Colors helps highlight important things to the server owner (and anyone helping them out with issues). I think all log messages should allow color. Feel free to contest me on this if you still prefer otherwise.

MrIvanPlays: I agree that all messages should allow color but go find out a method of logging exceptions with a colored message.

lokka30: We can definitely suspend the idea for later – if we ever find the need to use colored messages for the exception error logging method, we can just add in the feature then.

lokka30 (approved):
Looks great to me. 👍🏻

PR #214, opened 2022-10-31
EconomyTransactionType#SET

### LINKED ISSUE #198: Set/reset balance operations should use a new transaction type `SET`
### User Story

Suggested by `Hai_tun#8468` and myself on the [ArcanePlugins Discord Server](https://www.discord.io/arcaneplugins).

### Is your feature request related to a problem?

_No response_

### Describe the solution you'd like.

Instead of set/resets not being a transaction, they should be a transaction of a new Type named `SET`.

### Describe alternatives you've considered.

Keeping them as current: not transactions.

### Additional information

_No response_

MrIvanPlays:
Closes #198 

Removed `setBalance` method in order to force the use of `doTransaction`
Didn't directly push because needs discussion about shall we create a `resetBalance` method with a reason and a importance.

### DIFF
--- a/api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
+++ b/api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
@@ -246,21 +246,48 @@ public interface Account {
     /**
      * Reset the {@code Account} balance to its starting amount.
      *
-     * <p>Certain implementations, such as the {@link PlayerAccount}, may default to non-zero starting balances.
+     * <p>Certain implementations, such as the {@link PlayerAccount}, may default to non-zero
+     * starting balances.
      *
      * @param initiator    the one who initiated the transaction
      * @param currency     the {@link Currency} of the balance being reset
+     * @param importance   the reset transaction importance
      * @param subscription the {@link EconomySubscriber} accepting the new balance
-     * @see PlayerAccount#resetBalance(EconomyTransactionInitiator, Currency, EconomySubscriber)
-     * @since v1.0.0
+     * @see #resetBalance(EconomyTransactionInitiator, Currency, EconomyTransactionImportance, String, EconomySubscriber)
+     * @since v2.0.0
      */
     default void resetBalance(
             @NotNull EconomyTransactionInitiator<?> initiator,
             @NotNull Currency currency,
+            @NotNull EconomyTransactionImportance importance,
+            @NotNull EconomySubscriber<BigDecimal> subscription
+    ) {
+        resetBalance(initiator, currency, importance, null, subscription);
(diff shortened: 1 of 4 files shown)


api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
```diff
@@ -276,25 +252,26 @@ void doTransaction(
...
-            }
-        });
+        doTransaction(EconomyTransaction
+                .newBuilder()
+                .withCurrency(currency)
+                .withInitiator(initiator)
+                .withTransactionAmount(BigDecimal.ZERO)
+                .withReason("reset")
+                .withImportance(EconomyTransactionImportance.NORMAL)
+                .withTransactionType(EconomyTransactionType.SET)
```
lokka30, line 273: As mentioned in the OP, we need to have these values specified by parameters in the reset method. Re-posting it here so it is not lost amongst the forthcoming reviews and comments.

api/src/main/java/me/lokka30/treasury/api/economy/transaction/EconomyTransactionType.java
```diff
@@ -25,6 +25,13 @@ public enum EconomyTransactionType {
      *
      * @since v1.0.0
      */
-    WITHDRAWAL
+    WITHDRAWAL,
+
+    /**
+     * The Account's new balance is set. Can be more or less than their previous balance.
```
lokka30, line 31:
```suggestion
     * The Account's balance is adjusted to a specified value.
```

lokka30 (changes requested):
Looks great so far!

api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
MrIvanPlays: in that case we would be probably better off removing the `resetBalance` method too

lokka30: I'm lenient to go that route, however, one thing that does strike me is how the resetBalance method creates a standardized way of resetting the balance for a particular account type. For instance, PlayerAccount balances are reset to the starting balance of a currency, whilst non-PlayerAccount balances are reset to zero.

MrIvanPlays: okay lets just make the importance and reason configurable.

lokka30 (approved):
Sweet! 👍🏻

Jikoo (approved):
The reset method seems like a useful shortcut and demonstrates proper usage, +1 to that.

PR #220, opened 2022-11-03
Make migration working

### LINKED ISSUE #217: Update migration code to use futures
See the `//TODO` comments in `dev/2.0.0` (migration package)

MrIvanPlays:
Closes #217


MrIvanPlays:
@Jikoo can you review please so we can continue working on v2?

MrIvanPlays:
I've thought of migrating the logic from that long-ass chained countdown latch spaghetti to my new [Process](https://github.com/MrIvanPlays/Process) library. Opinions wanted.

core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/economy/migrate/MigrationEconomy.java
```diff
@@ -69,29 +75,20 @@ public boolean isPrimary() {
             }
 
             @Override
-            public boolean supportsNegativeBalances() {
-                return false; // todo: ?????
+            public TriState supportsNegativeBalances() {
+                return TriState.UNSPECIFIED;
```
Jikoo, line 79: What's wrong with returning a value indicating that the migration economy doesn't support negative balances? It doesn't support balances at all, but that's beside the point. Changing the entire API for what is just an ideally very temporary placeholder that denies access feels weird.

core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/economy/migrate/AccountMigrator.java
```diff
@@ -5,161 +5,137 @@
...
-            @Override
-            public void phaseFail(@NotNull final EconomyException exception) {
-                migration.debug(() -> getErrorLog(fromAccount.getIdentifier(), exception));
-                fromCurrencies.completeExceptionally(exception);
+        CountDownLatch latch = new CountDownLatch(1);
+        String accountId = fromAccount.getIdentifier();
+        fromAccount.retrieveHeldCurrencies().thenAccept(resp -> {
+            if (!resp.isSuccessful()) {
+                migration.debug(() -> getErrorLog(accountId, resp.getFailureReason()));
+                return;
```
Jikoo, line 51: Doesn't this result in migration never completing because the latch will never count down? Would prefer to see the migration "complete" with failures rather than leave it stalled indefinitely.

core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/economy/migrate/EconomyMigrateSub.java
```diff
@@ -202,108 +204,125 @@ public List<String> complete(
...
-                    public void phaseFail(@NotNull EconomyException exception) {
-                        migration.debug(() -> migrator.getBulkFailLog(exception));
-                    }
-                }
-        );
+        CountDownLatch latch = new CountDownLatch(1);
+        migrator.requestAccountIds(migration.from().get()).thenAccept(resp -> {
+            if (!resp.isSuccessful()) {
+                migration.debug(() -> migrator.getBulkFailLog(resp.getFailureReason()));
+                return;
```
Jikoo, line 237: Same here, migration will hang indefinitely.

Jikoo (commented):
Overall looks good, just a couple concerns about migration hanging.

core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/economy/migrate/MigrationEconomy.java
MrIvanPlays: we're in the 2.0.0 branch. supportsNegativeBalances method was added in 2.0.0 . we can break it as we wish. what's wrong with returning TrIState rather than a boolean?

core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/economy/migrate/AccountMigrator.java
MrIvanPlays: see how easy it is to miss such stuff? what's your opinion on migrating that whole latch shit I made to my new library [Process](https://github.com/MrIvanPlays/Process) ?

core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/economy/migrate/EconomyMigrateSub.java
MrIvanPlays: see comment above

core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/economy/migrate/MigrationEconomy.java
Jikoo: I don't understand why it needs to be an ambiguous value at all. The migration economy does not support negative balances. Granted, it also does not support balances, but that's not the point really - there is not negative balance support, so a return value of false is accurate.

core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/economy/migrate/AccountMigrator.java
Jikoo: If it makes life easier, all for it. Migration is more than a little messy.

MrIvanPlays:
I did a explanation in ArcanePlugins discord when I made the [Process](https://github.com/MrIvanPlays/Process) library and I'm gonna do a quick rundown here and what is the potential advantage: 

Current code relies on at least 1 countdown latch per account (2 latches if the account is a non player one), and a countdown latch for all the account types migrated. Combined with the `.thenApply`/`.thenCompose` calls on all the futures and stuff is currently a literal spaghetti. Another disadvantage is that the code waits for 1 account to finish migration before proceeding to another account.

With Process we can implement a simple `Process` class for player accounts and non player accounts, make `x` amount of instances of that implementation and run them in parallel and have it wait (either sync or async) for all of them to finish. Process uses only 1 (!) countdown latch per `runProcesses` call, reducing the spaghetti down to nonexistent.

To add to this, since all the migration code is running async I'm pretty sure we're good to replace all the `.thenApply`/`.thenCompose` spaghetti with simple `.get`/`.join` calls. My tests conclude they do not swallow exceptions.

Jikoo:
That sounds good to me. I think the old Phaser code did run pretty parallel, but it was also an absolute nightmare of back and forth. Honestly, anything to make migration cleaner.

MrIvanPlays:
Alright. I think in about 2 hours ill be done implementing it.

MrIvanPlays:
taking longer than expected, i have to do a modification to my library

api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
```diff
@@ -413,6 +419,49 @@ CompletableFuture<Response<TriState>> setPermission(
...
+                        for (UUID uuid : ids) {
+                            CompletableFuture<Response<Map<AccountPermission, TriState>>> future = fifo.poll();
+                            if (future == null) {
+                                throw new RuntimeException("Queue ended but there are still ids?");
+                            }
+                            // safe to call .join - since it's all been done
+                            Response<Map<AccountPermission, TriState>> response = future.join();
+                            if (!response.isSuccessful()) {
+                                // TODO: this is swallowing potential failures - METHOD FOR
+                                //  LOGGING NEEDED
```
MrIvanPlays, line 455: any ideas??

core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/economy/migrate/PlayerAccountMigrationProcess.java
```diff
@@ -0,0 +1,156 @@
...
+        }
+
+        for (String id : heldCurrenciesResp.getResult()) {
+            // find currency in the new provider
+            Optional<Currency> currencyOpt = this.migration.to().findCurrency(id);
+            Currency currency;
+            if (currencyOpt.isPresent()) {
+                currency = currencyOpt.get();
+            } else {
+                // TODO: perhaps create the currency in the new provider?
```
MrIvanPlays, line 69: shall we implement this? the original v1 code did not.

core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/economy/migrate/PlayerAccountMigrationProcess.java
```diff
@@ -0,0 +1,156 @@
...
+        Response<Collection<String>> heldCurrenciesResp = fromAccount
+                .retrieveHeldCurrencies()
+                .get();
+        if (!handleUnsuccessfulResponse(heldCurrenciesResp)) {
+            return;
+        }
+
+        for (String id : heldCurrenciesResp.getResult()) {
+            // find currency in the new provider
+            Optional<Currency> currencyOpt = this.migration.to().findCurrency(id);
```
MrIvanPlays, line 64: im unsure which service provider was the original implementation was trying to get here via the service provider. @Jikoo is the `to` provider usage here correct?

MrIvanPlays (commented):
I need some specific thoughts on these.

MrIvanPlays:
[img](https://img.mrivanplays.com/ihegrxtsop.png) 😁
Special thanks to: BetterEconomy v2.6-ivanized
running the version of treasury on the latest commit on this branch.
[img1](https://img.mrivanplays.com/idfolaolpu.png)

api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
Jikoo: If logging errors here is difficult and the default implementation is not advisable due to how many requests it makes, should we just not have a default implementation? Alternately, would failing to fetch data for one UUID be considered a full failure because it is not a complete set of permission mappings?

core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/economy/migrate/PlayerAccountMigrationProcess.java
Jikoo: That looks correct. I think in the initial v1 migrator we pre-mapped currencies for migration, but this should work fine.

core/src/main/java/me/lokka30/treasury/plugin/core/command/subcommand/economy/migrate/PlayerAccountMigrationProcess.java
Jikoo: I think there initially wasn't a way to create a new currency via the API, but I could be wrong. The v1 migrator pre-mapped currencies, and that would imo be an appropriate way to create currencies if we do plan to map them. That way all currencies, not just ones currently held by players, would be migrated.

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
```diff
@@ -94,7 +95,7 @@ public interface Currency {
      * @return boolean value
      * @since v2.0.0
      */
-    boolean supportsNegativeBalances();
+    TriState supportsNegativeBalances();
```
Jikoo, line 98: Was hoping to revert this to a boolean because ambiguous responses don't really make sense, but it's probably fine as-is either way. Definitely a minor nitpick.

Jikoo (approved):
Looks good to me overall.
If we do want to migrate currencies I think that should likely be done prior to accounts to reduce duplicate code and prevent redundant failures if creation doesn't work.

api/src/main/java/me/lokka30/treasury/api/economy/currency/Currency.java
lokka30: Agreed, this is a yes or no response.

MrIvanPlays:
Also a quick clarification: I'm not merging this until we test it.

MrIvanPlays:
Waiting for the day
When @Jikoo will review this

MrIvanPlays:
Tested and works. I'm still waiting @Jikoo

MrIvanPlays:
image

PR #228, opened 2022-11-17
Add missing nullability annotations

### LINKED ISSUE #221: Add nullability annotations on methods
since v2 migrated from subscribers to futures, we forgot to add `@NotNull` annotations on the return values of methods changed.

MrIvanPlays:
Also added a test for them and decided to improve the POM

Closes #221


lokka30 (approved):
Brilliant, thanks for adding the annotation test too

PR #229, opened 2022-11-22
Implement Permissions API

MrIvanPlays:
Currently, this lays the ground of a Permissions & Chat API for Treasury. 
I'm opening this not just for discussion, but to implement sane ideas suggested in this PR.


api/src/main/java/me/lokka30/treasury/api/permission/Context.java
```diff
@@ -0,0 +1,47 @@
...
+        this.world = world;
+    }
+
+    @NotNull
+    public String context() {
+        return this.context;
+    }
+
+    @NotNull
+    public TriState isWorldContext() {
```
lokka30, line 43: Why is this a TriState? What does `UNKNOWN` represent?

What if we want a proxy server context? The current impl only supports world (and global) contexts.

api/src/main/java/me/lokka30/treasury/api/permission/NodeHolder.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.Collection;
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+import me.lokka30.treasury.api.common.misc.TriState;
+import me.lokka30.treasury.api.common.response.Response;
+import org.jetbrains.annotations.NotNull;
+
+public interface NodeHolder {
+
+    @NotNull UUID getUniqueId();
```
lokka30, line 16: NonPlayerAccounts in economies do not use UUIDs. We need some consistency.

api/src/main/java/me/lokka30/treasury/api/permission/NodeHolder.java
```diff
@@ -0,0 +1,30 @@
...
+
+    @NotNull CompletableFuture<Response<Collection<Node<?>>>> allNodes();
```
lokka30, line 18:
```suggestion
    @NotNull CompletableFuture<Response<Collection<Node<?>>>> getNodes();
```

api/src/main/java/me/lokka30/treasury/api/permission/PermissionsProvider.java
```diff
@@ -0,0 +1,19 @@
...
+
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+import me.lokka30.treasury.api.common.response.Response;
+import org.jetbrains.annotations.NotNull;
+
+public interface PermissionsProvider {
+
+    @NotNull
+    NodeFactory obtainNodeFactory();
```
lokka30, line 15: Why `obtain`? Could `nodeFactory()` cut it? :)

lokka30 (commented):
Thanks Ivan! I've taken a look at what you've currently implemented, and left some comments.

api/src/main/java/me/lokka30/treasury/api/permission/Context.java
MrIvanPlays: `UNKNOWN` represents the global context.

> What if we want a proxy server context? The current impl only supports world (and global) contexts.

world and server are kinda the same thing. it'd be world if it's about a server, and server if it's about a proxy. I guess we need another name for `world`, something between `world` and `server`

api/src/main/java/me/lokka30/treasury/api/permission/NodeHolder.java
MrIvanPlays: breaks consistency with `allNodesWithContext` and `allNodesWithType`

MrIvanPlays:
there's another problem we need to discuss: that's about prefix and suffix nodes. their data type is a component. but the problem is that spigot and bungeecord utilize bungeecord-chat whilst any other server/proxy implementation (sponge, velocity, paper, etc.) utilizes adventure. if we opt in for bungeecord-chat components, then we need to download bungeecord-chat for the platforms where they don't have it and translate bungeecord-chat -> adventure, or the reverse, if we opt in for adventure components, then we need to download adventure for the platforms where they don't have it and translate adventure -> bungeecord-chat. a solution i thought of is create our own chat api, but that's basically just unnecessary reinvention of the wheel.

api/src/main/java/me/lokka30/treasury/api/permission/PermissionsProvider.java
lokka30: I think `nodeFactory` is best

MrNemo64:
Sorry for avandoning Treasury for so long. I just dont have time to keep up. But I saw that you're asking for help.
I haven't got time to check what ideas you have proposed but:
> solution i thought of is create our own chat api, but that's basically just unnecessary reinvention of the wheel.
Are these two systems too diferent from each other? Couldn't you just abstract the common/neccessary functionality and use the abstraction on the api, but on the code have it call the system of the platform?
Sorry if this doesn't help much, but I don't have time to keep up with all the progress

MrIvanPlays:
> Are these two systems too diferent from each other?

Yes. Completely.

> Couldn't you just abstract the common/neccessary functionality and use the abstraction on the api, but on the code have it call the system of the platform?

No.

api/src/main/java/me/lokka30/treasury/api/permission/context/ContextType.java
```diff
@@ -0,0 +1,42 @@
...
+        boolean hasSpace = false;
+        for (int i = 0; i < value.length(); i++) {
+            if (Character.isWhitespace(value.charAt(i))) {
+                hasSpace = true;
+                break;
+            }
+        }
+        if (hasSpace) {
+            throw new IllegalArgumentException("Invalid value: value should not contain spaces!");
+        }
```
montlikadani, line 39:
```suggestion
        if (value.indexOf(' ') != -1) {
            throw new IllegalArgumentException("Invalid value: value should not contain spaces!");
        }
```
Better readability and optimized.

api/src/main/java/me/lokka30/treasury/api/permission/node/Node.java
```diff
@@ -0,0 +1,28 @@
...
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public interface Node<Data> {
+
+    @NotNull NodeType<Data> nodeType();
+
+    @NotNull Context context();
+
+    @Nullable Integer weight();
```
montlikadani, line 18: I would rather use the primitive instead of the wrapper type of int and set the default value to `-1`, or we could make a WeightBuilder/NodeWeightBuilder to separate the actual weight and the permission node context.

MrIvanPlays: I first had it with the primitive, but then told myself how are we gonna define no weight presence? Using -1 is not really the way, because some permission plugins may opt in to use such weight.

montlikadani: true, just thought if it could work

api/src/main/java/me/lokka30/treasury/api/permission/node/holder/NodeHolder.java
```diff
@@ -0,0 +1,43 @@
...
+
+    /**
+     * Retrieves all the {@link Context contexts} this {@code NodeHolder} holds {@link Node nodes}
+     * under.
+     *
+     * @return
+     */
+    @NotNull CompletableFuture<Response<Collection<NodeHolder>>> retrieveHeldContexts();
+
+    @NotNull <Data> CompletableFuture<Response<TriState>> insertOrModifyNode(@NotNull Node<Data> node);
```
montlikadani, line 41: Could also add a method to remove node(s)

montlikadani:
> a solution i thought of is create our own chat api

The only solution this ^. We could make a simplifed and small Component/NameComponent for these prefix/suffix to retrieve and change things. Or just a plain String as the last resort.

MrIvanPlays:
> > a solution i thought of is create our own chat api
> 
> The only solution this ^. We could make a simplifed and small Component/NameComponent for these prefix/suffix to retrieve and change things. Or just a plain String as the last resort.

I really don't want to think this is the only solution. An idea just passed through my mind and that is create a ComponentCreator that will spit out the component json, then consumers and implementors can shove that in bungee-chat or adventure and get a proper component.

MrIvanPlays:
@Jikoo any thoughts/ideas?

Jikoo (commented):
I agree on the prefix/suffix nodes being a huge pain. There hasn't been a golden standard for how chat prefixes work. Putting the prefix and suffix into the permission plugin does sound like a good idea because it enables plugins not directly dependent on the chat plugin to get the actual components, but LP's meta system is relatively unique. Basic permission plugins do not necessarily support arbitrary text attachments, it's usually handled on the chat plugin side. Maybe we would be better served by a "separate" API for player metadata including prefixes and suffixes instead?

Re: components: Agreed, writing our own chat component API sounds awful. What if we have a wrapper that presents both the implementation-dependent real component as a generic Object and a `toJSON` method? Still leaves Spigot/Paper dual support a bit of a mess for people trying to use the API, but it solves any Adventure-supporting platform pretty neatly without require the extra back-and-forth.
Alternately, just use Adventure and use the Libraries feature for Spigot specifically, though that does mandate min version 1.18.

Would also like to point out that even LP does not serve components AFAIK. It serves raw text that the chat plugin then parses.

api/src/main/java/me/lokka30/treasury/api/permission/PermissionsProvider.java
```diff
@@ -0,0 +1,28 @@
...
+
+    @NotNull NodeFactory nodeFactory();
+
+    @NotNull CompletableFuture<Response<NodeHolderPlayer>> retrievePlayerNodeHolder(@NotNull UUID uuid);
+
+    @NotNull CompletableFuture<Response<NodeHolder>> retrieveNodeHolder(@NotNull String identifier);
+
+    @NotNull CompletableFuture<Response<Collection<UUID>>> allPlayerNodeHolders();
+
+    @NotNull CompletableFuture<Response<Collection<String>>> allNodeHolders();
```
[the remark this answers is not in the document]
MrIvanPlays: Question is what is a node holder? I dont think non-player is suitable here.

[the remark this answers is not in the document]
lokka30: Let's discuss that topic [here](https://github.com/ArcanePlugins/Treasury/discussions/233).

Jikoo:
I think we have to have a concept of groups and inheritance if we are going to allow write actions. What happens if I try to remove a node that is declared via a parent group? Is there an error? Does it succeed with no effect because the node is still declared via the parent? Is the parent group affected instead of the individual user? Should the permission implementation negate the node or explicitly declare it undefined for the user? In its current state, the API does not really define the actual behavior to expect, which to me means it isn't good.

As far as groups go, the only major downside I can think of is that they add implementation complexity. However, pretty much every established permissions plugin has a concept of groups because manually managing individual players' permissions would be an administrative nightmare. The other fiddly bit is that a group is effectively the same as a `NodeHolder`, it just happens to be an abstract construct used for the sake of holding multiple nodes, potentially in a hierarchy. I think the best solution to this would be to have an empty interface `NodeGroup` subclassing `NodeHolder`.

If Treasury wants groups, we need effective nodes (what we already have), explicit nodes, and inheritance (`CompletableFuture<Collection<NodeGroup>> NodeHolder#getParents`?). It would also be nice to have a way to fetch groups, although I'm not sure how full a system we want this to be. Do we want to then allow group creation and destruction? It makes sense if we're allowing general manipulation. There's a lot to consider.

Jikoo:
> > What happens if I try to remove a node that is declared via a parent group?
> 
> It shall be removed only for that node holder it has been removed from and it should stay in the parent one. E.g. group `a` has `/treasury info`, player `b` has group `a` as parent, removing node `treasury.command.treasury.info` from player `b` results in player `b` not having access to `/treasury info`, but all other players (player `c` and player `d` for example) that also inherit group `a` have access to `/treasury info`.
> 
> > Is there an error? Does it succeed with no effect because the node is still declared via the parent? Is the parent group affected instead of the individual user?
> 
> No to all of these.

Right, right, but the point is that all are interpretations that are "valid" to some degree. More to highlight the fact that the behavior is very ambiguous without a concept of inheritance. If I say "remove X permission from user" I would expect the end result to be that they no longer have the permission set. If I can't check where they're getting it from, I can't take any action to remove it correctly.

> Nodes are nodes. Why shall we break them down? "explicit nodes, inheritance nodes" sounds like bullshit and bloat to me.

Explicit nodes would be nodes that are actually declared in a state for the holder. I do think it's important to give access to that information if we are going to permit manipulation of nodes' states. Not important for most cases, but anyone doing manipulation of nodes is likely going to care about the actual internal states. See the questions I posed for possible results. What if one of those results is actually what I want to have happen? How do I make sure that I've actually identified the true source of the node for removal/negation?
Additionally, if we are not offering explicit states, permission nodes should not be a `TriState` at all. Undeclared states are good in the context of "is a permission explicitly set" but it's annoying in the context of "do I have permission to perform this action" because then I need to perform additional logic (presumably platform-specific, too) to figure out what undefined behavior is. The permission plugin (should have) baked this information already to speed up queries, the user manually recalculating it is either inaccurate or an expensive waste.

How do I determine a permission's state for a context? Ex.: `remotebank.use` is true globally.  `remotebank.use` is false in the world "creative." How do I query that node? Do I fetch all nodes for that context and manually try to find it? That then loops back to me having to find out what undefined behavior is, but specifically for that context. It also seems inefficient when most plugins would bake states to be queriable by key, then context.

MrIvanPlays:
you are free to push to this branch. I am unable to fully understand what you're trying to state but if you implement it then I will understand what you're talking about. I approve any changes, this is a WIP after all, we can always change stuff. my understanding of contexts and nodes have never been complete.

Jikoo:
All right, will do. May end up just creating a separate branch because I imagine I'll end up changing a lot and I don't want to hold you back if it takes me a forever.

MrIvanPlays:
Another idea for components: Have a `ComponentAccessor` service in the API that gives essential methods for creating components and outputs the raw JSON. That service will be implemented and maintained by the treasury plugin.

MrIvanPlays:
> Another idea for components: Have a `ComponentAccessor` service in the API that gives essential methods for creating components and outputs the raw JSON. That service will be implemented and maintained by the treasury plugin. 

No responses here, any opinions on this?

lokka30:
So long it is very small and easy to maintain.

api/src/main/java/me/lokka30/treasury/api/permission/context/ContextSet.java
```diff
@@ -0,0 +1,75 @@
...
+        this.contexts = contexts;
+    }
+
+    @NotNull
+    public Optional<Context> getContext(@NotNull String contextKey) {
+        Objects.requireNonNull(contextKey, "contextKey");
+        return Optional.ofNullable(this.contexts.get(contextKey));
+    }
+
+    public void registerContext(@NotNull Context context) {
```
A248, line 51: This method seems out-of-place. I thought ContextSet was immutable, hence `ContextSet.empty()` and your use of Collections.emptyMap(), which is immutable.

It's best to decide whether you want ContextSet to be mutable and maintain consistency in your decision afterward.

api/src/main/java/me/lokka30/treasury/api/permission/node/Node.java
A248: I've another idea. Why not use `int` anyway? We can allow permission plugins to return the default weight for a permission group if no weight is explicitly defined.

This way, client plugins can use and rely on the permissions API by working with effective weights, rather than merely the weights configured. If weights aren't configured for a specific permissions group, the default weight may be returned.

api/src/main/java/me/lokka30/treasury/api/permission/node/holder/NodeHolderPlayer.java
```diff
@@ -0,0 +1,57 @@
...
+     * {@link me.lokka30.treasury.api.permission.context.Context contexts} defined in the
+     * specified context set's conditions are met.
+     *
+     * @param contextSet context set to probe
+     * @return applies or not.
+     */
+    boolean contextSetApplies(@NotNull ContextSet contextSet);
+
+    @NotNull
+    default CompletableFuture<Response<TriState>> hasPermission(@NotNull String nodeKey) {
```
A248, line 41: I'm wondering whether using CompletableFuture here is the best approach. Rather than have a data request be performed every time `hasPermission` is called, why not create a method to pre-load user permission data? The client plugin may request permission data for a user; once loaded, this permission data is in-memory so `hasPermission` calls using the data object return synchronously.  I've never used the LuckPerms API but I believe it allows something like this.

A248 (commented):
Here are my ideas so far

api/src/main/java/me/lokka30/treasury/api/permission/node/holder/NodeHolderPlayer.java
lokka30: I think that should not be in the hands of the API, and rather, the permissions service provider. Regardless, it would not be ideal to add a synchronous cache-assuming `hasPermission` method since that will be abused by developers instead of using the CF version, even when they are not certain that permissions have been cached for a user.

A248: What do you think of this idea? We might design the API so that it is impossible to use a cache-backed `hasPermission` without first loading such cache.

For example, a `loadUser(UUID)` method returns a `CompletableFuture<LoadedUser>`. Then, the synchronous `LoadedUser#hasPermission` allows checking permissions synchronously.

lokka30: The API could handle caching, though it is imperative that permission service provider developers fully understand that they must update the cache of a permission holder when some relevant part of their context changes.

For instance, if permission X is only given to player Y in world Z, and player Y changes to world W, then the cache should be updated primarily to ensure that permission X only remains allocated if the player remains in world Z.   

Another consideration to make is that the rest of the APIs do not offer inbuilt caching. Ideally, we keep things consistent between our APIs.

Overall, I'm uncertain about that idea, and more inclined to think it would not be the best route forward.

api/src/main/java/me/lokka30/treasury/api/permission/node/Node.java
MrIvanPlays: I'm really 50/50 on this. More opinions wanted.

api/src/main/java/me/lokka30/treasury/api/permission/context/ContextSet.java
MrIvanPlays: Yeah I also think I shall transform this immutable.

api/src/main/java/me/lokka30/treasury/api/permission/node/holder/NodeHolderPlayer.java
MrIvanPlays: Current design allows for caching. Changing the design to something like `loadUser` will often lead to confusion between both permission plugins and plugins which use the api. 

Consider this: plugin x uses the api and calls `loadUser` each time they need it, which causes multiple database queries to be performed, because developer y of plugin x doesn't know the implementation cache is being updated (if it really is, because permission plugin z may choose not to update it). The current design significantly reduces database queries, because one may not need the whole information about the account.

api/src/main/java/me/lokka30/treasury/api/permission/node/Node.java
lokka30: I agree with A2 here.

MrIvanPlays: @Jikoo @MrNemo64 thoughts?

MrIvanPlays:
@Jikoo when will you be able to implement your ideas here?

Jikoo:
Probably not for at least a month or two based on how things are looking - my GPU is failing, which in turn makes pretty much everything quite frustrating. I was very spoiled by dual 1440p monitors, a single at 1024x768 is practically unusable.

Please don't hold back on my account.

MrIvanPlays:
This is not gonna push back the 2.0.0 release anymore. We will release this when it's ready.

api/src/main/java/me/lokka30/treasury/api/permission/node/Node.java
MrIvanPlays: @Jikoo @MrNemo64 thoughts?

lokka30: I still believe a primitive is definitely way the go. Using the wrapper type here would unnecessarily complicate the permission weights system.

Jikoo: I think that a primitive is also the way to go - no weight should likely be handled the same way a weight of 0 would be for any calculations, so we're just forcing anyone doing those calculations to unwrap it.

MrIvanPlays:
@Jikoo please implement your ideas whenever you can

Jikoo:
This is the next major item on my list, but my development time is currently being devoted to a plugin of mine that has apparently been partially broken since like 1.14 due to changes that weren't listed on the MC wiki (whoops).

MrIvanPlays:
> This is the next major item on my list, but my development time is currently being devoted to a plugin of mine that has apparently been partially broken since like 1.14 due to changes that weren't listed on the MC wiki (whoops).

okay, I just reworked nodes btw

MrIvanPlays:
it looks like to me there was 0 time to work on this :( . Anyone still interested in giving his ideas into writing this API?

Issue #225, opened 2022-11-09
Support more platforms


MrIvanPlays:
- [x] Sponge Support
- [x] Minestom Support

lokka30:
Is it worth delaying 2.0 for platform support?

Advantage: We are able to revise the API to support those platforms if required, instead of having to either abandon the idea of supporting that platform, make an unfavourable workaround, or embarrassingly break the API again. 

Disadvantage: v2 is delayed considerably longer than hoped.

What do you think? 🙂

MrIvanPlays:
yes it is worth

PR #230, opened 2022-11-23
Support Sponge 

MrIvanPlays:
This PR partly solves #225, by adding SpongeAPI support to Treasury.

### DIFF
--- a/platform/sponge/src/main/java/me/lokka30/treasury/plugin/sponge/SpongeTreasuryPlugin.java
+++ b/platform/sponge/src/main/java/me/lokka30/treasury/plugin/sponge/SpongeTreasuryPlugin.java
@@ -15,7 +15,9 @@ import me.lokka30.treasury.plugin.core.config.settings.Settings;
 import me.lokka30.treasury.plugin.core.logging.Logger;
 import me.lokka30.treasury.plugin.core.schedule.Scheduler;
 import me.lokka30.treasury.plugin.core.utils.PluginVersion;
+import net.kyori.adventure.text.Component;
 import net.kyori.adventure.text.serializer.legacy.LegacyComponentSerializer;
+import net.kyori.adventure.text.serializer.plain.PlainTextComponentSerializer;
 import org.jetbrains.annotations.NotNull;
 import org.spongepowered.api.Sponge;
 import org.spongepowered.api.scheduler.Task;
@@ -29,13 +31,15 @@ public class SpongeTreasuryPlugin extends TreasuryPlugin implements Logger, Sche
     private Messages messages;
     private Settings settings;
 
-    private final File messagesFile;
-    private final File settingsFile;
+    private final File messagesFile, settingsFile;
+    private final Path pluginsFolder;
 
     public SpongeTreasuryPlugin(TreasurySponge plugin) {
         this.plugin = Objects.requireNonNull(plugin, "plugin");
         messagesFile = new File(plugin.getDataDir().toFile(), "messages.yml");
         settingsFile = new File(plugin.getDataDir().toFile(), "settings.yml");
+
+        pluginsFolder = plugin.getDataDir().getParent().getParent().resolve("mods");
 
         // todo: somehow parse back the container#metadata#version back into a string
         // put a valid version for now so we can test whether the plugin works
@@ -54,7 +58,7 @@ public class SpongeTreasuryPlugin extends TreasuryPlugin implements Logger, Sche
 
     @Override
     public @NotNull Path pluginsFolder() {
-        return plugin.getDataDir().getParent();
+        return pluginsFolder;
     }
 
     @Override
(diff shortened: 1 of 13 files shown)


platform/sponge/src/main/java/me/lokka30/treasury/plugin/sponge/SpongeTreasuryPlugin.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    @Override
+    public @NotNull Platform platform() {
+        return Platform.SPONGE;
+    }
+
+    @Override
+    public @NotNull Path pluginsFolder() {
+        return null; // TODO
```
montlikadani, line 52: To retrieve the plugins directory (mods) you need `Sponge.configManager().pluginConfig(pluginContainer).directory().getParent()`

MrIvanPlays: Thanks!

montlikadani:
Also can you list out a todo list what need to be implemented? Or do you want everything the same that is in Bukkit?

MrIvanPlays:
@montlikadani I'm new to the sponge API, can you please do a full review and tell me what to change? Thanks in advance

> Also can you list out a todo list what need to be implemented? Or do you want everything the same that is in Bukkit?

I was writing this when you commented. But basically:
	- A command source/sender/executor wrapper
	- Command implementation (did something idk if it's right, but I read sponge supports brigadier with its literals and arguments and I'd like such implementation rather Command.Raw)
	- Parse back the version to a string so our PluginVersion class can handle it
	- Review & Test.

A PR to this branch will be greatly appreciated too, and you will be credited accordingly.

MrIvanPlays:
Oh and also, our core implementation had `Scheduler#runSync`. I removed it because I couldn't find a way to run tasks in Sponge on the main thread. I'd like this method back, even though it is unused. It may be needed in the future.

montlikadani:
> Oh and also, our core implementation had `Scheduler#runSync`. I removed it because I couldn't find a way to run tasks in Sponge on the main thread. I'd like this method back, even though it is unused. It may be needed in the future.

To execute in synchronous thread, you can use
```java
Sponge.server().scheduler().submit(Task.builder().plugin(pluginContainer).execute(() -> {

}).build());
```

Where `org.spongepowered.api.scheduler.Task#builder` provides you many scheduling related methods. `execute` method can also accept Runnable and Consumer functions too.

MrIvanPlays:
Thanks! Can you please review the current code and point where I can improve stuff?

MrIvanPlays:
@montlikadani How do I get the plugin jar in Sponge? 
Also I recommend you to get in the ArcanePlugins discord so we can chat there: https://discord.gg/HqZwdcJ

montlikadani:
> - A command source/sender/executor wrapper

In order to make Parameterized command, you need to build a command using `Command#builder`. I made a simple command builder with one parameter:
```java
private Parameter.Value<ServerPlayer> playerArgumentValue;

@Listener
public void onRegisterCommands(RegisterCommandEvent<Command.Parameterized> event) {
	Command.Parameterized exampleCommand = Command.builder().permission("name.permission")
		.shortDescription(Component.text("This is a short description of this command argument"))
		.addParameters(playerArgumentValue = Parameter.player().key("player")
			.optional().requiredPermission("name.argument.player").build()
		.completer((context, currentInput) -> currentInput.isEmpty()
			? Arrays.asList(CommandCompletion.of("completableArgumentName"))
			: Collections.emptyList()).build()).executor(this::processExample).build();

		event.register(plugin,
				Command.builder().addChild(exampleCommand, "example").terminal(true)
				.executor(this::processExample).build(), "maincommand", "mc");
	}
```
For the first look this looks so many lines and codes. But this is how Sponge implemented their command execution.

This is an example of how you can make a parameterized command. To simplify what is this exactly:
`onRegisterCommands` this is a method name in which you can register and add as many commands as you want, this method will be called when the server state is ready to register commands. This is why `@Listener` annotation added to this method to mark that we need this method to be called. 

`Command.Parameterized` is an interface which allows you to create arguments for your main command. There is no limit of how many of parameters you have, you just need to build them. The main command here is the `maincommand` or `mc` which is the short aliase. You can add more labels/aliases if you want.

`addParameters` method adds parameters to your argument like as I created an example `player` which awaits the player name and requires `name.argument.player` permission for the executor (sender). You can add more parameters to this argument if you want.

You can see that `playerArgumentValue` is cached outside of this method, because we need to use it later for checking the command contexts when performing the command.

`completer` is the same like in Bukkit, tab-completer in which you can add parameters to complete them on pressing the tab key.

Will do the others ...

platform/sponge/src/main/java/me/lokka30/treasury/plugin/sponge/TreasurySponge.java
```diff
@@ -0,0 +1,102 @@
...
+    TreasurySponge(PluginContainer container, Logger logger) {
+        this.logger = logger;
+        this.container = container;
+    }
+
+    private SpongeTreasuryPlugin treasuryPlugin;
+    private TreasuryCommand command;
+
+    @Listener
+    public void onStart(StartedEngineEvent<Server> event) {
```
montlikadani, line 48: There are other interfaces which will be called on different stages (or lifecycles) like this. This is called when the server is ready to use, so in your current implementation you initializes everything at once which not really harmful, but I would preferable extract this into two different stages.

In my case I would do this:
- Creating instances of classes, creating/loading configs would get into `StartingEngineEvent`
- If you want listeners to be registered, you can do it in LoadedGameEvent or in this method StartedEngineEvent, the order doesn't matter

Also, if you implements reload command, you can also add a new listener method with `RefreshGameEvent` parameter, but this event is only called when you execute `/sponge plugins reload pluginName` from game.

platform/sponge/src/main/java/me/lokka30/treasury/plugin/sponge/TreasuryCommand.java
```diff
@@ -0,0 +1,160 @@
...
+    private List<CommandCompletion> complete(CommandContext context, String[] args) {
+        List<String> completions = this.base.complete(this.sources.obtainSource(context
+                .cause()
+                .audience()), "treasury", args);
+
+        if (completions == null || completions.isEmpty()) {
+            return Collections.emptyList();
+        }
+
+        List<CommandCompletion> ret = new ArrayList<>();
```
montlikadani, line 153: Set an initial size for this list
```suggestion
        List<CommandCompletion> ret = new ArrayList<>(completions.size());
```

montlikadani (approved):
Overall LGTM, if it worked then we're ready unless there is something that we need to do.

MrIvanPlays:
> Overall LGTM, if it worked then we're ready unless there is something that we need to do.

Tested on spongevanilla, if you can please test it on spongeforge or whatever was it called. I am unable to right now.

montlikadani:
SpongeForge implements SpongeAPI same as on SpongeVanilla, so it should work.

MrIvanPlays:
> SpongeForge implements SpongeAPI same as on SpongeVanilla, so it should work.

Then we're good. I just need @lokka30's approval and I'm merging it.

lokka30 (approved):
LGTM 👍🏻 

Thanks so much @MrIvanPlays and @montlikadani.

PR #239, opened 2023-01-03
Add NamespacedKey

MrIvanPlays:


api/src/main/java/me/lokka30/treasury/api/common/NamespacedKey.java
```diff
@@ -0,0 +1,160 @@
...
+     * @return new namespaced key
+     * @throws IllegalArgumentException if the inputted string is not correctly formatted, if the
+     *                                  inputted string contains any whitespaces, or if
+     *                                  the namespace is "treasury". The "treasury" namespace is
+     *                                  reserved for the Treasury plugin.
+     */
+    @NotNull
+    public static NamespacedKey fromString(@NotNull String namespacedKey) {
+        Objects.requireNonNull(namespacedKey, "namespacedKey");
+        validateWhitespaces(namespacedKey, "A namespaced key string cannot contain spaces");
```
lokka30, line 55: we need to assert that there is only a single colon character in the string

api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
```diff
@@ -38,14 +38,6 @@
...
 public interface Account {
 
-    /**
-     * Gets the string-based unique identifier for this account.
-     *
-     * @return The String unique identifier for this account.
-     * @since v1.0.0
-     */
-    @NotNull String getIdentifier();
-
```
lokka30, line 48: Should we undo this change and instead make `getIdentifier()` use the `Object` type? Every account needs a way to identify itself so it would be nice if we could somehow force that in the interface. No worries if this is not feasible.

api/src/main/java/me/lokka30/treasury/api/economy/account/AccountData.java
```diff
@@ -74,14 +75,14 @@ public boolean isPlayerAccount() {
...
-     * will return the identifier of the account. Otherwise, an empty
+     * will return the {@link NamespacedKey} identifier of the account. Otherwise, an empty
      * {@link Optional} will be returned.
      *
      * @return identifier or empty optional if the data is a player account data
      * @since 2.0.0
      */
     @NotNull
-    public Optional<String> getIdentifier() {
+    public Optional<NamespacedKey> getIdentifier() {
```
lokka30, line 85: It may be better to have the methods named as so:
- `Optional<NamespacedKey> getNonPlayerIdentifier()`
- `Optional<UUID> getPlayerIdentifier()`

lokka30 (changes requested):
Thanks as always Ivan :)

api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
MrIvanPlays: the raw account type shouldn't be really used imo, any specific interactions should be done via the appropriate interface type.

lokka30: we should make sure to clarify that the account interface does not support custom implementations

api/src/main/java/me/lokka30/treasury/api/common/NamespacedKey.java
```diff
@@ -0,0 +1,171 @@
...
+            }
+            if (c == ':') {
+                columnCharOccurrences++;
+                if (columnCharOccurrences > 1) {
+                    break;
+                }
+            }
+        }
+        if (columnCharOccurrences > 1 || columnCharOccurrences == 0) {
+            throw new IllegalArgumentException("namespacedKey string should contain a single ':'");
```
lokka30, line 68:
```suggestion
            throw new IllegalArgumentException("namespacedKey string should only contain a single colon (':') character");
```

api/src/main/java/me/lokka30/treasury/api/common/NamespacedKey.java
```diff
@@ -0,0 +1,171 @@
...
+        for (char c : namespacedKey.toCharArray()) {
+            if (Character.isWhitespace(c)) {
+                throw new IllegalArgumentException("A namespaced key string cannot contain spaces");
+            }
+            if (c == ':') {
+                columnCharOccurrences++;
+                if (columnCharOccurrences > 1) {
+                    break;
+                }
+            }
```
lokka30, line 65: I wonder if it would be clear if we moved this check down just below where we do the string split and check that `parts.length == 2`

lokka30 (changes requested):

lokka30 (approved):

PR #247, opened 2023-02-01
Implement proposed changes in #246

### LINKED ISSUE #246: General Economy API Improvement Proposals
### `EconomyFailureReason`

- [x] **Complete removal of this enum**
  - Ivan and I believe this enum should be removed. See [here](https://github.com/ArcanePlugins/Treasury/issues/246#issuecomment-1409853985) for reasoning.
  - [x] *Approvals: @lokka30, @MrIvanPlays*
  - [x] Resolved

### `Currency`
- [x] `supportsNegativeBalances`
  - This method should be removed.
  - We discussed earlier that Treasury does not care about negative balances, that is all handled by the economy provider. If the economy provider thinks that a given account is not capable of receiving a given transaction, then they will just prevent the transaction from going through.
  - [x] *Approvals: @lokka30, @MrIvanPlays*
  - [x] Resolved

- [x] `getStartingBalance(UUID)`:
  - I am thinking of moving this to `Account` and instead making it `getStartingBalance()`.
  - This means that each account can be given a different starting balance if the economy provider wants that.
  - The current implementation allows only players to have custom starting balances.
  - It also seems quite weird to have a context-centric method (dealing with player UUIDs) in an environment which does not really directly with any economy accounts.
  - [x] *Approvals: @lokka30, @MrIvanPlays*
  - [x] Resolved

### EconomyProvider

- [x] `retrieveAllAccountsPlayerIsMemberOf` and `retrieveAllAccountsPlayerHasPermission`
  - These need to return account objects, not IDs.
  - theoretically, the economy provider can override these methods and have more optimised approach since they have direct access to their database, but in practice, nobody is going to do this
  - people want the accounts - they are going to retrieve those anyways - so we need to just return those
  - as an API, Treasury should not need to care how this method is implemented so long it gets the job done as the method contracts specify 
  - [x] *Approvals: @lokka30, @MrIvanPlays*
  - [x] Resolved

### Miscellaneous

- [ ] In many javadocs we tell implementers to return something in the case of a failure occurring, when `null` is expected (as per `Response.failure`). If something fails, we want to know why it failed, not some useless result value. The javadocs should be updated so that a `result` is not expected if the given `Response` results in a failure.
  - [x] *Approvals: @lokka30, @MrIvanPlays*
  - [ ] Resolved

***

### Miscellaneous To-do's

- [x] In `EconomyProvider#unregisterCurrency`'s javadoc, I need to fix a spelling mistake, "`IF`" instead of "`If`".

MrIvanPlays:
Aims to fix #246


api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
```diff
@@ -73,6 +73,15 @@ public interface Account {
     @NotNull
     CompletableFuture<Response<BigDecimal>> retrieveBalance(@NotNull Currency currency);
 
+    /**
+     * Gets the starting balance of this {@code Account} for the specified {@link Currency currency}
```
lokka30, line 77:
```suggestion
     * Gets the starting balance of this account for the specified {@link Currency currency}.
```

MrIvanPlays: `{@code Account}` looks better. prove me wrong

api/src/main/java/me/lokka30/treasury/api/economy/account/PlayerAccount.java
```diff
@@ -172,7 +178,7 @@ default CompletableFuture<Response<BigDecimal>> resetBalance(
                 .newBuilder()
                 .withCurrency(currency)
                 .withInitiator(initiator)
-                .withTransactionAmount(currency.getStartingBalance(getUniqueId()))
+                .withTransactionAmount(this.getStartingBalance(currency))
```
lokka30, line 181:
```suggestion
                .withTransactionAmount(getStartingBalance(currency))
```

MrIvanPlays: Why?

lokka30 (approved):
Changes so far look great, I've left some comments. Not ready to merge this PR until we have discussed all current (and future) proposed changes in #246.

api/src/main/java/me/lokka30/treasury/api/economy/account/PlayerAccount.java
lokka30: I reiterate the same question for prefixing the method with `this`. It's clearly calling the `getStartingBalance` method within itself.

api/src/main/java/me/lokka30/treasury/api/economy/account/Account.java
lokka30: It's not code, neither it's not a specific key to look for in the config. Would you prefer:

```java
{@link Account account}
```

?

MrIvanPlays: img

See the difference? It gets a different font.

api/src/main/java/me/lokka30/treasury/api/economy/account/PlayerAccount.java
MrIvanPlays: I just realized we can remove the method from PlayerAccount and transfer it to just Account. `this.` is a Java feature that is to better illustrate which method/field _exactly_ you are calling.

lokka30: Oops, yes, it was meant to be in Account. Can't believe I missed that in my initial review. Regardless, I don't care much about using `this` to die on a hill lol, but I don't think it's necessary at all here.

lokka30 (approved):

PR #266, opened 2023-03-21
Wrap Treasury's Economy API to Sponge's Economy API inside treasury-sponge

MrIvanPlays:


platform/sponge/src/main/java/me/lokka30/treasury/plugin/sponge/apiimpl/economy/SpongeToTreasuryCurrencyImpl.java
```diff
@@ -0,0 +1,120 @@
...
+        // No starting balance for migrated currencies sponge -> treasury
+        return BigDecimal.ZERO;
+    }
+
+    @Override
+    public @NotNull BigDecimal getConversionRate() {
+        // This is kinda sad imo, currencies coming from the Sponge API will not be able to get
+        // properly converted in between (Sponge API currency -> Sponge API currency) when utilised
+        // by the users of the treasury api.
+        return BigDecimal.ONE;
```
Jikoo, line 93: Is there anything we can do on our end to alleviate this? From a quick glance over the Sponge economy API it doesn't look like they have any concept of conversion at all. Would it be better to make this return ZERO and disallow conversion entirely, or am I misremembering what that does and that would make it convert into infinite money?

platform/sponge/src/main/java/me/lokka30/treasury/plugin/sponge/apiimpl/economy/UniqueAccountImpl.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package me.lokka30.treasury.plugin.sponge.apiimpl.economy;
+
+import java.util.UUID;
+import me.lokka30.treasury.api.economy.EconomyProvider;
+import me.lokka30.treasury.api.economy.account.PlayerAccount;
+import org.spongepowered.api.service.economy.account.UniqueAccount;
+
+public class UniqueAccountImpl extends VirtualAccountImpl implements UniqueAccount {
```
Jikoo, line 12: All Treasury-backed UniqueAccounts being VirtualAccounts seems to violate the docs of VirtualAccount:
> Represents an Account which is not tied to a particular User.
>
> Examples of virtual accounts include:
> * A global server account
> * A 'bank account', shared among multiple users
> * A account for a non-player Entity.

Should probably have VirtualAccountImpl become an `AbstractAccountImpl implements Account` and then extend it to VirtualAccountImpl/UniqueAccountImpl.

Jikoo (changes requested):
Overall looks good, one change and one question. There were a couple spots where I was a little confused, but then I realized that Sponge only allows for one economy... so either Treasury provides the economy or it wraps a Sponge economy, and there are no overlapping states where both cases could be true.

platform/sponge/src/main/java/me/lokka30/treasury/plugin/sponge/apiimpl/economy/SpongeToTreasuryCurrencyImpl.java
MrIvanPlays: Valid point here, will change it to zero

platform/sponge/src/main/java/me/lokka30/treasury/plugin/sponge/apiimpl/economy/UniqueAccountImpl.java
MrIvanPlays: The difference between a `UniqueAccount` and `VirtualAccount` is that `UniqueAccount` features a `UUID` as its identifier, nothing else. Hence, in order to remove some implementation burden we can safely extend the `VirtualAccount` implementation.

Jikoo: I get that this is one fewer implementation to maintain, but it's still a violation of the docs. VirtualAccount is an empty interface extending Account specifically for non-player account types.

MrIvanPlays: .... ok I'll do an abstract account implementation

Jikoo (approved):
LGTM 👍

MrIvanPlays:
Thanks. Waiting on @lokka30

Jikoo (approved):

A248:
I don't mean to be too much of a contrarian. However, I want to know about how this PR interacts with economy plugins providing their own implementation of the Sponge economy API.

Let's say I write a multi-platform economy plugin implementing the Treasury API. I support Sponge. On Sponge, moreover, I want to provide my custom implementation of the Sponge Economy API that handles matters slightly differently from the Treasury adapter implementation (contained in this PR). For example, maybe I want to provide a customizable or more sophisticated way to handle Sponge requests that would otherwise cause me to use `CompletableFuture.join()`.

This is not a theoretical concern, at least I do not think so. The reason is because of lessons learned maintaining my plugin LibertyBans. In LibertyBans, I support some Sponge ban management-related APIs in a way that is slightly different than merely wrapping my own API. Therefore, I imagine someone writing an economy plugin might have similar analogous reasons for wanting to implement the Sponge Economy API separately.

MrIvanPlays:
@A248 There's a config option to disable Treasury wrapping the Treasury Economy API onto Sponge's Economy API. I don't know how will Sponge react to 2 `ProvideServiceEvent` listeners that provide an economy implementation, so I don't know what will happen if both this wrap is enabled whilst another plugin is providing.

lokka30 (commented):
Upon a glance over the code, things seem excellent. Before I grant this a seal of approval, I would like to make sure that Treasury's player and non player account types are comfortably interchangeable with Sponge's unique and virtual account types. Thank you to Ivan and others who have contributed to this PR for the fantastic work, I hope the Sponge community will be able to more easily adopt and incorporate Treasury with this adapter layer.

MrIvanPlays:
> Upon a glance over the code, things seem excellent. Before I grant this a seal of approval, I would like to make sure that Treasury's player and non player account types are comfortably interchangeable with Sponge's unique and virtual account types. Thank you to Ivan and others who have contributed to this PR for the fantastic work, I hope the Sponge community will be able to more easily adopt and incorporate Treasury with this adapter layer.

Yes they are. A wrapped by Treasury `VirutalAccount`'s `String` identifier can be easily converted back into a `NamespacedKey` using the `fromString` method, from which you can retrieve a `NonPlayerAccount`. As for `UniqueAccount`s and `PlayerAccount`s, the identifier is a `UUID`, so it's the same business there.

lokka30:
Lovely docs over at Sponge [here](https://docs.spongepowered.org/stable/en/plugin/economy/basics.html), yep it'll be perfect. Thanks so much Ivan, A2 and Jikoo. Squash and merge when you're ready 👍🏻

lokka30 (approved):

PR #270, opened 2023-04-04
Replace EconomyTransactionInitiator with Cause

MrIvanPlays:
I think this is needed as even though I am the one who created the EconomyTransactionInitiator I was confused by it. The new Cause interface is abstracted out from the economy api, which allows it's usage into other future apis (permissions api for example), and it has been made easier to use, with also adding one more type of initiators: accounts.

Before:
```java
EconomyTransactionInitiator<UUID> initiator = EconomyTransactionInitiator.createInitiator(Type.PLAYER, playerUuid);
// ...
```

After:
```java
Cause.Player cause = Cause.player(playerUuid);
// ...
```

This also allows to have more OOP than previously in this section, whilst also having a future-proof interface where if we need a new cause we don't need to break stuff.


Jikoo (approved):
LGTM 👍 Clean and expressive

MrIvanPlays:
I've decided to simplify it by making every PlayerAccount and NonPlayerAccount a Cause. @Jikoo if you'd like you can do another review here.

Jikoo (approved):
Looks good, making Accounts Causes is a clever way to simplify things. I do like that it automatically clarifies the potential confusion between obtaining a Cause for a PlayerAccount of `Cause#player` vs `Cause#account`.

MrIvanPlays:
Waiting on @lokka30

lokka30:
This is an excellent idea with equally great execution.

One thing I would like clarified is Cause#account's name, I would expect a method of this name to return something that is for generic Accounts however it is intended for Non-Player accounts.

Should we rename `account()` to `nonPlayer()` so it's a little more consistent with `player()`? Note that it won't be fully consistent since `player` references a UUID rather than a PlayerAccount.

Cheers. This is an excellent improvement to the API.

lokka30 (changes requested):
see above

PR #285, opened 2023-07-04
Always check for classes on BukkitVendor

HSGamer:
This ultimately fixes a case where it fails to check Folia vendor if the server is running on a fork of Folia (like [Kaiiju](https://github.com/KaiijuMC/Kaiiju)).
This also changed the target classes based on [PaperLib](https://github.com/PaperMC/PaperLib/blob/69e76fc9036e7b5f2c967bd21db31043724ef9e3/src/main/java/io/papermc/lib/PaperLib.java#L37-L45) and my own [HSCore](https://github.com/HSGamer/HSCore/blob/master/hscore-bukkit/hscore-bukkit-folia/src/main/java/me/hsgamer/hscore/bukkit/folia/FoliaChecker.java) (which is based on a discussion on PaperMC's Discord server)

P/S: Yes, they did port some classes from Folia back to Paper, but only [the Scheduler & Owner Region API](https://github.com/PaperMC/Paper/blob/master/patches/server/0972-Folia-scheduler-and-owned-region-API.patch). Therefore, the old class checks are still relevant.

### DIFF
--- a/platform/bukkit/src/main/java/me/lokka30/treasury/plugin/bukkit/vendor/BukkitVendor.java
+++ b/platform/bukkit/src/main/java/me/lokka30/treasury/plugin/bukkit/vendor/BukkitVendor.java
@@ -5,7 +5,6 @@
 package me.lokka30.treasury.plugin.bukkit.vendor;
 
 import me.lokka30.treasury.plugin.core.Platform;
-import org.bukkit.Bukkit;
 
 /**
  * Represents a handler for determining on what server vendor we're running. This is in order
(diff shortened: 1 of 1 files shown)


MrIvanPlays (approved):

platform/bukkit/src/main/java/me/lokka30/treasury/plugin/bukkit/vendor/BukkitVendor.java
```diff
@@ -43,19 +42,31 @@ public static Platform getPlatformClass() {
...
+    private static boolean hasClass(String... classNames) {
+        for (String className : classNames) {
+            try {
+                Class.forName(className);
+                return true;
+            } catch (ClassNotFoundException ignored) {
+            }
+        }
+        return false;
+    }
```
lokka30, line 60: I'm being pedantic here - may be worth renaming this method to `hasAnyClasspath`.

MrIvanPlays: Doesn't matter, if it works, it works. It's also treasury bukkit plugin internal stuff.

lokka30: Resolved in https://github.com/ArcanePlugins/Treasury/commit/698658395d54b215e759a611f919828db413fc19

